PR #2, opened 2018-10-20
Task/api 86 H2 configuration for testing with temporary databases

bschofield-va:
- Updates secrets to discover spring passwords when used in scripts
- Updates to allowed patterns so that variables can be used, e.g. `password=$WOW_SECRET`
- H2 database library is automatically added to the target directory for easy access
- `run-local.sh` test script for launching `ids` using an ephemeral H2 database


bschofield-va:
@eclendenning-va Let's get this merged first thing in the morning so it's behind us

PR #4, opened 2018-10-23
Task/api 83 ids smoke tests

bschofield-va:
This is an early look at the integration test machine. The intention is to incorporate
- Beacon
- Em's client-type-per-service Beacon experiment
- Agent K
- Ian's Categories Agent K experiment
- Automatic process management for local testing

The ultimate goal is to have one machine that can be used to test locally running services or services in any environment including prod.

The follow command will build the services, then launch them and run integration tests, finally tearing down the services when finished.
```
mvn clean install -Dsentinel=LOCAL
```


bschofield-va:
This looks bigger than it is since it includes https://github.com/department-of-veterans-affairs/health-apis/pull/5

PR #5, opened 2018-10-23
Task/api 87 ids registration

bschofield-va:
Registration support for identity service part 1
    - ids/README now describes the behavior of the identity service
    - ids/run-local.sh now builds before running
    - new ids/dmeo.sh script to show off a little functionality
    - Added validation annotations to IDs model
    - basic registration support for ids, but no error handling
    - validation errors returned as 400.


bschofield-va:
We'll add some integration tests in `83`

PR #7, opened 2018-10-23
Task/api 89 lookup support

bschofield-va:
Adds the remaining error checks for lookup.


bschofield-va:
@eclendenning-va nice find, fixed

PR #14, opened 2018-10-29
Task/api 105 argo api skeleton

eclendenning-va:
Patient model with test coverage and Swagger generation.


argonaut-api/src/main/java/gov/va/api/health/argonaut/api/Patient.java
```diff
@@ -0,0 +1,93 @@
...
+  @Valid List<ContactPoint> telecom;
+
+  @Pattern(regexp = Fhir.CODE)
+  @NotBlank
+  Gender gender;
+
+  @Pattern(regexp = Fhir.DATE)
+  String birthDate;
+
+  // TODO mutualExclusive validator
```
ilaflamme-va: We need to remember to revisit this(Story/Task if there isn't one already). It's not essential for this merge, but we will want this functionality soon.

PR #31, opened 2018-11-09
Task/api 184 ethnicity extension validation

eclendenning-va:
https://vasdvp.atlassian.net/browse/API-184
- Added validation for patients with an argo-ethnicity extension
- Created tests and sample data to ensure validation works

### DIFF
--- a/argonaut-api/src/main/java/gov/va/api/health/argonaut/api/Patient.java
+++ b/argonaut-api/src/main/java/gov/va/api/health/argonaut/api/Patient.java
@@ -1,13 +1,18 @@
 package gov.va.api.health.argonaut.api;
 
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import gov.va.api.health.argonaut.api.validation.RelatedFields;
 import gov.va.api.health.argonaut.api.validation.ZeroOrOneOf;
 import io.swagger.v3.oas.annotations.media.Schema;
 import java.util.List;
+import java.util.Optional;
 import javax.validation.Valid;
+import javax.validation.constraints.AssertTrue;
 import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotEmpty;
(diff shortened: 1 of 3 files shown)


argonaut-api/src/main/java/gov/va/api/health/argonaut/api/Patient.java
```diff
@@ -87,4 +89,39 @@
     other,
     unknown
   }
+
+  @JsonIgnore
+  @AssertTrue(message = "Argo-Ethnicity extension is not valid")
+  private boolean isValidEthnicityExtension() {
+    if (extension != null) {
```
bschofield-va: - Refactor to remove the heavy nesting by returning early when possible
- Move variable declarations closer to their use
- The count comparison logic is currently looking for what is _not_ valid. The method can be simplified a little by looking for what _is_ valid.

```
if (extension == null) { 
  return true;
}
Optional<Ext> ethnicityExtention = extension.stream()
    ...
if (!ethnicityExtention.isPresent()) {
  return true;
}
int ombExtensionCount = 0;
int textExtensionCount = 0; 
for ( ... ) {
  ...
}
return ombExtensionCount < 1 && textExtensionCount == 1;

File: argonaut-api/src/test/java/gov/va/api/health/argonaut/api/validation/ArgoEthnicityExtensionValidationTest.java [hunk 2]
```diff
@@ -0,0 +1,53 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Set;
+import javax.validation.ConstraintViolation;
+import javax.validation.Validation;
+import javax.validation.ValidatorFactory;
+import org.junit.Test;
+
+public class ArgoEthnicityExtensionValidationTest {
```
bschofield-va: Nice test! Clear names, clean code. Love it 🌮

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/Patient.java
eclendenning-va: Polish applied!

PR #37, opened 2018-11-14
Task/api 270 patient search bundler

bschofield-va:
Adds patient search support to argonaut
Updated generated swagger to have search parameters
Updates cdw-schema to latest version (with repackaging and renames)


argonaut/src/test/java/gov/va/api/health/argonaut/service/controller/BundlerTest.java
```diff
@@ -0,0 +1,118 @@
...
+  Bundler bundler;
+
+  @Before
+  public void _init() {
+    MockitoAnnotations.initMocks(this);
+    bundler = new Bundler(links);
+  }
+
+  @Test
+  public void deleteme() {
```
ilaflamme-va: This should likely be renamed to something identifiable, like bundlerBuildsGenericallyTypedBundle.

PR #54, opened 2018-11-19
Task/api 323 diagnostic report model

ilaflamme-va:


argonaut-api/src/main/java/gov/va/api/health/argonaut/api/ArgonautService.java
```diff
@@ -29,6 +30,73 @@
...
+  )
+  @POST
+  @Consumes("application/fhir+json")
+  @Path("DiagnosticReport/$validate")
+  @ApiResponse(
+    responseCode = "200",
+    description = "Record found",
+    content =
+        @Content(
+          mediaType = "application/fhir+json",
```
jgarnhart-va: Would it be worth creating a global constant for "application/fhir+json" that could be used by this and all other resources?

ilaflamme-va: Good idea. We'll make this happen once all of these ongoing PRs settle.

PR #57, opened 2018-11-20
Task/api 331 immunization model

eclendenning-va:
Resource: http://www.fhir.org/guides/argonaut/r2/StructureDefinition-argo-immunization.html
Story: https://vasdvp.atlassian.net/browse/API-331
- Built out Immunization Model
- Created Immunization sample data
- Added test support for model and bundle


argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Immunization.java
```diff
@@ -0,0 +1,207 @@
...
+
+  @Pattern(regexp = Fhir.CODE)
+  String language;
+
+  @Valid Narrative text;
+  @Valid List<SimpleResource> contained;
+  @Valid List<Extension> extension;
+  @Valid List<Extension> modifierExtension;
+  @Valid List<Identifier> identifier;
+  @NotNull Status status;
```
esubocz-va: needs @Valid

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Immunization.java
```diff
@@ -0,0 +1,207 @@
...
+    @NotNull
+    @Min(1)
+    Integer doseSequence;
+
+    String description;
+    @Valid Reference authority;
+    String series;
+
+    @Min(1)
+    Integer seriesDoses;
```
esubocz-va: Card is 0...1, why @Min(1)?

eclendenning-va: @Min constrains the value of the Integer. seriesDoses is a postiveInt which is defined as any positive integer (e.g. > 0) .

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Immunization.java
eclendenning-va: Nice Catch!

argonaut-api/src/test/java/gov/va/api/health/argonaut/api/resources/ImmunizationTest.java
```diff
@@ -0,0 +1,63 @@
...
+import gov.va.api.health.argonaut.api.samples.SampleImmunizations;
+import java.util.Collections;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.Test;
+
+@Slf4j
+public class ImmunizationTest {
+
+  private final SampleImmunizations immunizationData = SampleImmunizations.get();
+  private final SampleDataTypes dataTypes = SampleDataTypes.get();
```
esubocz-va: With new structure, these aren't both necessary. A declaration like:
private final SampleImmunizations data = SampleImmunizations.get(); will work. & references to each can be replaced by data. (you can use other resource tests for reference)

eclendenning-va: Fixed!

PR #64, opened 2018-11-26
Task/api 336 allergy intolerance read and sentinel

esubocz-va:
Added support for Allergy Intolerance read in Argonaut, and added sentinel tests.
Related story:  https://vasdvp.atlassian.net/browse/API-336


bschofield-va:
I've addressed these issues and this is ready to go.

PR #72, opened 2018-11-28
Task/api 385 encounter model

eclendenning-va:
Related Story: https://vasdvp.atlassian.net/browse/API-385
Related Resource: http://www.hl7.org/fhir/DSTU2/encounter.html

Added:
- Model
- Sample data 
- Tests

### DIFF
--- a/argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Encounter.java
+++ b/argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Encounter.java
@@ -1,0 +1,231 @@
+package gov.va.api.health.argonaut.api.resources;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import gov.va.api.health.argonaut.api.Fhir;
+import gov.va.api.health.argonaut.api.bundle.AbstractBundle;
+import gov.va.api.health.argonaut.api.bundle.AbstractEntry;
+import gov.va.api.health.argonaut.api.bundle.BundleLink;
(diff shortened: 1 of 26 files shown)


argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Encounter.java
```diff
@@ -0,0 +1,231 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@Schema(description = "http://www.hl7.org/fhir/DSTU2/encounter.html")
+public class Encounter implements DomainResource {
+
+  @Pattern(regexp = Fhir.ID)
+  String id;
+
+  @NotBlank String resourceType;
```
bschofield-va: Prefer this as the first field so it serializes to Json in the same order as the existing apps

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Encounter.java
```diff
@@ -0,0 +1,231 @@
...
+
+  @Valid List<Extension> extension;
+
+  @Valid List<Identifier> identifier;
+
+  @NotNull Encounter.Status status;
+
+  @Valid List<StatusHistory> statusHistory;
+
+  Encounter.Clazz clazz;
```
bschofield-va: I `clazz` was agreed upon, but now that I see it, I feel icky since it looks like reflection code... also, this will serialize as ` "clazz" : "inpatient"` which is not correct. Let's help Jackson out and rename the type field.

```
@JsonProperty("class")
EncounterClass encounterClass;

...

public enum EncounterClass {
  ambulatory
  ...
```

PR #83, opened 2018-12-04
Appointment Model

ilaflamme-va:

### DIFF
--- a/argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Appointment.java
+++ b/argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Appointment.java
@@ -1,0 +1,118 @@
+package gov.va.api.health.argonaut.api.resources;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import gov.va.api.health.argonaut.api.Fhir;
+import gov.va.api.health.argonaut.api.datatypes.CodeableConcept;
+import gov.va.api.health.argonaut.api.datatypes.Identifier;
+import gov.va.api.health.argonaut.api.datatypes.SimpleResource;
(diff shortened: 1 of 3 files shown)


argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Appointment.java
```diff
@@ -0,0 +1,171 @@
...
+  @Pattern(regexp = Fhir.CODE)
+  String language;
+
+  @Valid Narrative text;
+  @Valid List<SimpleResource> contained;
+  @Valid List<Extension> extension;
+  @Valid List<Extension> modifierExtension;
+  @Valid List<Identifier> identifier;
+
+  Status status;
```
esubocz-va: status is required. should be @NotBlank

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Appointment.java
```diff
@@ -0,0 +1,171 @@
...
+
+  String description;
+
+  @Pattern(regexp = Fhir.INSTANT)
+  String start;
+
+  @Pattern(regexp = Fhir.INSTANT)
+  String end;
+
+  int minutesDuration;
```
esubocz-va: positive int, should it have @Min ?

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Appointment.java
```diff
@@ -0,0 +1,171 @@
...
+
+  @Valid Reference slot;
```
esubocz-va: This should be a list according to spec. This change will require an edit to the sample data as well.

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Appointment.java
bschofield-va: It's also optional, should be Integer.

PR #84, opened 2018-12-04
Task/api 425 location model

ilaflamme-va:


argonaut-api/src/test/java/gov/va/api/health/argonaut/api/resources/LocationTest.java
```diff
@@ -0,0 +1,18 @@
...
+
+import gov.va.api.health.argonaut.api.samples.SampleLocations;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.Test;
+
+@Slf4j
+public class LocationTest {
+
+  private final SampleLocations data = SampleLocations.get();
+
```
eclendenning-va: needs the can build bundle test

PR #91, opened 2018-12-06
Argonaut fully qualified reference links

bschofield-va:
- Adds new Jackson ObjectMapper constomizations that will fully qualify references during serialization.
- Adds new `argonaut.base-path` configurable to account for lab environment where base path isn't `api`
- Changes the way bundle links are created such that they no longer take the full request URI, but rather just the resource name.

### DIFF
--- a/argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/allergyintolerance/AllergyIntoleranceController.java
+++ b/argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/allergyintolerance/AllergyIntoleranceController.java
@@ -16,7 +16,6 @@ import gov.va.api.health.argonaut.service.mranderson.client.Query;
 import gov.va.dvp.cdw.xsd.model.CdwAllergyIntolerance103Root;
 import java.util.Collections;
 import java.util.function.Function;
-import javax.servlet.http.HttpServletRequest;
 import lombok.AllArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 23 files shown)


argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/allergyintolerance/AllergyIntoleranceController.java
```diff
@@ -48,15 +47,11 @@
...
-      MultiValueMap<String, String> parameters,
-      int page,
-      int count,
-      HttpServletRequest servletRequest) {
+  private Bundle bundle(MultiValueMap<String, String> parameters, int page, int count) {
     CdwAllergyIntolerance103Root root = search(parameters);
     LinkConfig linkConfig =
         LinkConfig.builder()
-            .path(servletRequest.getRequestURI())
+            .path("AllergyIntolerance")
```
bschofield-va: @ilaflamme-va @eclendenning-va @esubocz-va 
Be aware of the changes to the controllers around paging links. The new lab environment uses a different base path: `/services/argonaut/v0` instead of `/api`. To deal with this, the Page Links now accepts an explicit configurable `argonaut.base-path` and expects the link config to just contain the resource name.

Changes
- Remove HttpServletRequest from controller
- Set _path_ to be the resource name
- Remove mock HttpServletRequest from controller unit tests

Note: You change this line in the bundle method, then tell Intellij to _Safe Delete_ the now unused parameter, then select the search methods in the dialog that pops up, it will automatically remove it everywhere and fix the unit tests. You'll still need to delete the mock HttpServletRequest field from the test, but it saves a lot of time.

PR #99, opened 2018-12-11
Argonaut now allows `0 <= _count <= 20` for all resources

bschofield-va:
Mock CDW driver will now loosely match queries when count == 0 or greater than 20 when patient or identifier values match.


argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/procedure/ProcedureController.java
```diff
@@ -100,7 +99,7 @@ private CdwProcedure101Root search(MultiValueMap<String, String> params) {
   public Procedure.Bundle searchById(
       @RequestParam("_id") String id,
       @RequestParam(value = "page", defaultValue = "1") @Min(1) int page,
-      @RequestParam(value = "_count", defaultValue = "1") @Min(1) @Max(20) int count,
+      @RequestParam(value = "_count", defaultValue = "1") @Min(0) int count,
       HttpServletRequest servletRequest) {
```
esubocz-va: This still uses the HttpServletRequest. Has been fixed in the resource sweep branch, but will cause a merge conflict again. (FYI)

PR #115, opened 2019-01-02
Bug/api 562 observation omit system when no code available

esubocz-va:
Nulls the system field in observation.valueQuantity.high, observation.valueQuantity.low, observation.referenceRange.high, and observation.referenceRange.low when no code is present.
Relevant bug: https://vasdvp.atlassian.net/browse/API-562

### DIFF
--- a/argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/observation/ObservationTransformer.java
+++ b/argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/observation/ObservationTransformer.java
@@ -38,6 +38,7 @@ import gov.va.dvp.cdw.xsd.model.CdwReference;
 import java.math.BigDecimal;
 import java.util.List;
 import javax.validation.constraints.NotNull;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.stereotype.Service;
 
 @Service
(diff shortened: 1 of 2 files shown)


argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/observation/ObservationTransformer.java
```diff
@@ -348,6 +354,16 @@ Quantity valueQuantity(CdwValueQuantity maybeCdw) {
             maybeCdw.getValue())) {
       return null;
     }
+    if (maybeCdw.getCode() == null) {
```
bschofield-va: Get rid of the if/else ... the only difference between the next 8 lines and the 9 below it is the _system_ setter.
```
Quantity.builder()
   .system(StringUtils.isBlank(cdw.getCode())?null:cdw.getSystem())
   ...
```

argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/observation/ObservationTransformer.java
```diff
@@ -289,6 +289,12 @@ SimpleQuantity referenceRangeQuantity(CdwObservationRefRangeQuantity maybeCdw) {
             maybeCdw.getSystem(), maybeCdw.getCode(), maybeCdw.getUnit(), maybeCdw.getValue())) {
       return null;
     }
+    if (maybeCdw.getCode() == null) {
+      return SimpleQuantity.builder()
```
bschofield-va: Kill the if/else... see my comments below.

esubocz-va:
@bschofield-va suggested edits, have been made.

PR #117, opened 2019-01-03
Task/api 593 Create DAR for Diagnostic Report Performer

esubocz-va:
Diagnostic Report has a required practitioner/organization reference field performer. A null value for this field will cause an invalid payload error. To prevent this, a data absent reason needs to be returned instead.
Related story: https://vasdvp.atlassian.net/browse/API-593

### DIFF
--- a/argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/diagnosticreport/DiagnosticReportTransformer.java
+++ b/argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/diagnosticreport/DiagnosticReportTransformer.java
@@ -7,8 +7,11 @@ import static gov.va.api.health.argonaut.service.controller.Transformers.convert
 import static gov.va.api.health.argonaut.service.controller.Transformers.ifPresent;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
+import gov.va.api.health.argonaut.api.DataAbsentReason;
+import gov.va.api.health.argonaut.api.DataAbsentReason.Reason;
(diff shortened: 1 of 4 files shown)


argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/diagnosticreport/DiagnosticReportTransformer.java
```diff
@@ -103,10 +106,31 @@ private DiagnosticReport diagnosticReport(CdwDiagnosticReport source) {
...
+        maybeSource,
+        source ->
+            Reference.builder()
+                .reference(source.getReference())
+                .display(source.getDisplay())
+                .build());
+  }
+
+  Extension performerExtenstion(CdwReference source) {
+    if (source == null) {
```
bschofield-va: If the reference is not null, but has null reference and display fields, we will produce and invalid response.
The checks for adding the performer reference and _performer extension need to be identical.

Recommend creating a helper ...
```
private boolean isUsable(Reference ref) { ... }
Reference performer(CdwReference maybeReference) {
   if (!isUsable(maybeReference)) {
     return null;
   }
  return convert( ... );
}

Extension performanceExtension(CdwReference maybeReference) {
   if (isUsable(maybeReference)) {
    return null;
   }
  return DataAbsentReason ...
}
```

esubocz-va: @bschofield-va suggested edits have been made.

PR #119, opened 2019-01-04
task/API-add-conformance-test-to-check-for-newlines-and-tabs

jjennings-va:
https://vasdvp.atlassian.net/browse/API-554

Added in check to see if conformance statement has any newline or tab characters


jjennings-va:
Wait to Merge: I am going to add in the Header 'Accept: application/json' to the test as this is what the customer uses and what has been a problem previously.

sentinel/src/main/java/gov/va/health/api/sentinel/ExpectedResponse.java
```diff
@@ -47,6 +48,15 @@ ExpectedResponse expect(int statusCode) {
     }
   }
 
+  /** Expect no newline or tab characters within JSON body. */
+  ExpectedResponse expectNoNewlineOrTabChars() {
```
bschofield-va: This is check is pretty specific to the conformance statement and should not be here. This should be moved to the one test body that needs it.  Furthermore, the `asString` is potentially expensive and should only be done once.

PR #120, opened 2019-01-06
Task/api 512 swagger example

jgarnhart-va:
https://vasdvp.atlassian.net/browse/API-512

When our Swagger yaml file gets built at build time, the resources now contain entries in their "@Schema" annotations for examples. Here is an example screen shot showing how the example gets included in the swagger.

swagger


jgarnhart-va:
@bschofield-va - I hear you about the missing examples. I was just going off the A/C of the story:

_A/C:
The developer portal page swagger shall have an example of each JArgonaut supported read & search._

jgarnhart-va:
@bschofield-va - Missing examples have been added as requested.

PR #122, opened 2019-01-09
Sentinel docker support

bschofield-va:
- Adds verifications to lab crawler tests
- Adds summary logging to lab crawler tests
- Adds docker packaging and utilities to execute tests
- New jar shader produces a _fat_ jar that includes all the dependencies. Some `test` scope dependencies were upgraded to `compile` to be included in the shaded jar
- New test jar which include test classes


sentinel/src/test/java/gov/va/health/api/sentinel/LabTest.java
```diff
@@ -8,34 +8,74 @@
...
+import lombok.extern.slf4j.Slf4j;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
 @Category(Lab.class)
+@Slf4j
 public class LabTest {
-  LabRobots robots = LabRobots.fromSystemProperties();
+
+  private static final String SAPIDER =
```
bschofield-va: @eclendenning-va

eclendenning-va: Love it!

PR #126, opened 2019-01-11
Support application/json+fhir

bschofield-va:
- Addresses error in Sentinel tests that were submitting a 'Content-Type' request header instead of 'Accept' for tests
- Addresses exception in Argonaut when Accept header with value application/json+fhir is requested

While fixing this, I verified the changes in Sentinel (adding the accept) line without the additional Spring Boot configuration indeed throws an exception.

@jgarnhart-va @aparcel-va This is part 1 of 3 issues that caused us downtime while updating prod.

### DIFF
--- a/argonaut/src/main/java/gov/va/api/health/argonaut/service/config/FhirMediaTypesConfig.java
+++ b/argonaut/src/main/java/gov/va/api/health/argonaut/service/config/FhirMediaTypesConfig.java
@@ -1,0 +1,43 @@
+package gov.va.api.health.argonaut.service.config;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.MediaType;
+import org.springframework.http.converter.HttpMessageConverter;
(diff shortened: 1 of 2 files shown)


argonaut/src/main/java/gov/va/api/health/argonaut/service/config/FhirMediaTypesConfig.java
```diff
@@ -0,0 +1,43 @@
...
+ * This configures spring to support additional, custom HTTP media types for FHIR. We need to
+ * support application/json, application/fhir+json, and application/json+fhir. Out of the box,
+ * Spring will support application/*+json, but special configuration is needed for
+ * application/json+fhir
+ */
+@Configuration
+public class FhirMediaTypesConfig implements WebMvcConfigurer {
+
+  private static final MediaType JSON_FHIR = MediaType.parseMediaType("application/json+fhir");
+
```
jgarnhart-va: Just a thought. We have this string alot in health-apis. Example:

```
content =
   27          @Content(
   28:           mediaType = "application/json+fhir",
   29            schema = @Schema(implementation = Patient.class)
   30          )
```

Wondering if this constant could be moved higher and used in place of strings.

PR #128, opened 2019-01-15
Task/api 656 handle two deceased values

eclendenning-va:
Related Story - https://vasdvp.atlassian.net/browse/API-656

- If both deceased values are populated, remove deceasedBoolean

### DIFF
--- a/argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/patient/PatientTransformer.java
+++ b/argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/patient/PatientTransformer.java
@@ -331,7 +331,7 @@ public class PatientTransformer implements PatientController.Transformer {
         .address(addresses(source.getAddresses()))
         .gender(gender(source.getGender()))
         .birthDate(asDateString(source.getBirthDate()))
-        .deceasedBoolean(deceasedBoolean(source.getDeceasedDateTime(),source.isDeceasedBoolean()))
+        .deceasedBoolean(deceasedBoolean(source.getDeceasedDateTime(), source.isDeceasedBoolean()))
         .deceasedDateTime(asDateTimeString(source.getDeceasedDateTime()))
         .maritalStatus(maritalStatus(source.getMaritalStatus()))
         .contact(contacts(source.getContacts()))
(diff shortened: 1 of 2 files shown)


argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/patient/PatientTransformer.java
```diff
@@ -330,7 +331,7 @@ private Patient patient(CdwPatient source) {
         .address(addresses(source.getAddresses()))
         .gender(gender(source.getGender()))
         .birthDate(asDateString(source.getBirthDate()))
-        .deceasedBoolean(source.isDeceasedBoolean())
+        .deceasedBoolean(deceasedBoolean(source.getDeceasedDateTime(), source.isDeceasedBoolean()))
```
bschofield-va: Could this be more plainly written?

```
.deceasedBoolean(source.getDeceasedTime()==null?source.getDeceasedBoolean():null)
```

Or if you wanted to be all fluently, you could create the opposite of `ifPresent`

```
.deceasedBoolean(unless(source.getDeceasedTime(),source::isDeceasedBoolean))
```

argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/patient/PatientTransformer.java
```diff
@@ -368,6 +369,13 @@ Gender gender(CdwAdministrativeGenderCodes source) {
     return ifPresent(source, gender -> EnumSearcher.of(Patient.Gender.class).find(gender.value()));
   }
 
+  Boolean deceasedBoolean(XMLGregorianCalendar deceasedDateTime, Boolean deceasedBoolean) {
+    if (deceasedDateTime != null) {
+      deceasedBoolean = null;
```
bschofield-va: As a general rule, reassigning input parameters is a bit of a code smell. For situations like this, I think this style is cleaner:
```
deceasedBoolean( DT dt, B b) {
   return dt == null?b:null;
}
```
or
```
if (dt == null) {
  return b;
}
return null;
```

PR #132, opened 2019-01-18
Task/API 668 Singular Specific Test Data to Set

esubocz-va:
In preparation of allowing sentinel to run in the PROD and QA environments, SystemDefinitions are added for each. Related Story: https://vasdvp.atlassian.net/browse/API-668

### DIFF
--- a/sentinel/src/main/java/gov/va/health/api/sentinel/SystemDefinitions.java
+++ b/sentinel/src/main/java/gov/va/health/api/sentinel/SystemDefinitions.java
@@ -83,19 +83,43 @@ public class SystemDefinitions {
                   .build())
           .build();
 
-  SystemDefinition PROD =
+  SystemDefinition prod =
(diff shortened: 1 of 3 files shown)


sentinel/src/main/java/gov/va/health/api/sentinel/SystemDefinitions.java
```diff
@@ -82,4 +82,97 @@
...
+          .build();
+
+  TestIds prodAndQaIds() {
+    return TestIds.builder()
+        .allergyIntolerance("800000001542")
+        .appointment("800295609816")
+        .condition("800000007353:P")
+        .diagnosticReport("800000005788:M")
+        .diagnosticReports(
+            DiagnosticReports.builder()
```
bschofield-va: DRY ... same values are used in local environment.

PR #133, opened 2019-01-22
logging and threading

ilaflamme-va:


service-auto-config/src/main/java/gov/va/api/health/autoconfig/configuration/SecureRestTemplateConfig.java
```diff
@@ -99,6 +99,12 @@ private ClientHttpRequestInterceptor loggingInterceptor() {
             "Response Body:\n{}",
             StreamUtils.copyToString(response.getBody(), StandardCharsets.UTF_8));
         log.error("----------------------------------------------------");
+      } else {
+        log.info(
+            "Response from {} {} is {}",
+            request.getMethod(),
+            request.getURI(),
+            response.getStatusCode());
```
jblefkove-va: Should this be combined with the log statement on line 85?

ilaflamme-va: I don't have the response until after the request is executed on line 86. So my thought was a pre-log and post-log.
I use the request.getMethod() and .getURI() as a unique identifier for the status code, in the case of a large volume of traffic in the logs.

jblefkove-va: You could move it into one log statement after the request is executed, before checking the error status. But I'm fine with explicit pre- and post-logging.

PR #134, opened 2019-01-22
Task/api 637 swagger availability

eclendenning-va:
Related Story : https://vasdvp.atlassian.net/browse/API-637

Looks for API documentation on :
- https://developer.va.gov/explore/health/docs/argonaut 
- https://dev-developer.va.gov/explore/health/docs/argonaut

### DIFF
--- a/sentinel/src/main/java/gov/va/health/api/sentinel/DevApiPortal.java
+++ b/sentinel/src/main/java/gov/va/health/api/sentinel/DevApiPortal.java
@@ -4,7 +4,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.File;
 import java.io.FileInputStream;
-import java.util.List;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import lombok.SneakyThrows;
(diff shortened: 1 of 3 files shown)


sentinel/src/main/java/gov/va/health/api/sentinel/DevApiPortal.java
```diff
@@ -0,0 +1,89 @@
...
+  public void initializeDriver(String url, int defaultTimeOutInSeconds) {
+    Config config = new Config(new File("config/lab.properties"));
+    ChromeOptions chromeOptions = new ChromeOptions();
+    chromeOptions.addArguments("--whitelisted-ips", "--disable-extensions", "--no-sandbox");
+    chromeOptions.setHeadless(config.headless());
+    if (StringUtils.isNotBlank(config.driver())) {
+      System.setProperty("webdriver.chrome.driver", config.driver());
+    }
+    driver = new ChromeDriver(chromeOptions);
+    wait = new WebDriverWait(driver, defaultTimeOutInSeconds);
```
jblefkove-va: I don't see the `wait` being used anywhere after it is declared here. Did I miss something?

sentinel/src/main/java/gov/va/health/api/sentinel/DevApiPortal.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private String valueOf(String name) {
+      String value = properties.getProperty(name, "");
+      assertThat(value).withFailMessage("System property %s must be specified.", name).isNotBlank();
+      return value;
+    }
+  }
+
+  /** Set up the chrome driver with the proper properties. */
+  public void initializeDriver(String url, int defaultTimeOutInSeconds) {
```
jblefkove-va: You don't need `default` in `defaultTimeOutInSeconds`. (It's not default since there's no way to override it.)

sentinel/src/main/java/gov/va/health/api/sentinel/DevApiPortal.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  public void quit() {
+    driver.quit();
+  }
+
+  /** Returns true if the element exists, and errors if not. */
+  public boolean isElementPresent(By by) {
+    driver.findElement(by);
+    return true;
```
jblefkove-va: If the method always returns `true`, you don't really need this return type. If the goal is to invoke `driver.findElement()` and verify that it doesn't throw an exception, this could be `void` instead. But in that case, this method is identical to `getElement()` above, so just use that one instead. The test can do ` assertThat(apiPortalPage.getElement(...)).isNotNull()`

PR #135, opened 2019-01-23
Task/api 672 extract tests into resource specific classes

esubocz-va:
ReadAndSearchIT has been split up into resource specific test classes, basic and advanced tests have been split into inner classes. Inner classes have been assigned junit categories to run only the appropriate tests given the specified environment.
Related story: https://vasdvp.atlassian.net/browse/API-672


sentinel/src/main/java/gov/va/health/api/sentinel/ServiceDefinition.java
```diff
@@ -13,13 +15,22 @@
...
-        .log()
-        .ifValidationFails();
+    RequestSpecification spec =
+        RestAssured.given()
+            .baseUri(url())
+            .port(port())
+            .relaxedHTTPSValidation()
+            .log()
+            .ifValidationFails()
+            .header("jargonaut", "true");
```
bschofield-va: Please change this to be specified from a system property `jargonaut`
* if not specified, omit header
* other specified, add header

PR #146, opened 2019-01-29
API-714 Handle medication code empty string.

jblefkove-va:
Replace `Transformers.allNull()` with `allBlank()`.

Update builder constructor for `Coding` to replace blank strings with nulls.


argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/medication/MedicationTransformer.java
```diff
@@ -72,15 +72,11 @@ CodeableConcept form(CdwCodeableConcept source) {
   }
 
   List<Coding> formCodings(List<CdwCoding> source) {
-    List<Coding> formCodings = convertAll(source, this::formCoding);
-    if (formCodings == null) {
-      return null;
-    }
-    return formCodings.isEmpty() ? null : formCodings;
+    return convertAll(source, this::formCoding);
```
bschofield-va: nice

PR #148, opened 2019-01-30
Task/api 734 crawl staging

esubocz-va:
We would like the web crawler to be able to point to the Staging environment. 
Related story: https://vasdvp.atlassian.net/browse/API-734


sentinel/src/main/java/gov/va/health/api/sentinel/crawler/UrlReplacementRequestQueue.java
```diff
@@ -0,0 +1,36 @@
...
+  RequestQueue requestQueue;
+
+  @Override
+  public void add(String url) {
+    if (isBlank(baseUrl)) {
+      throw new IllegalStateException("baseUrl not specified.");
+    }
+    if (isBlank(forceUrl)) {
+      throw new IllegalStateException("forceUrl not specified.");
+    }
```
jblefkove-va: Consider making the fields on this class `final`. Then, instead of checking that `baseUrl` and `forceUrl` are non-blank for every `add()`, you only need to do it once, when the object is constructed. (Generate an all-args constructor and put the lombok `@Builder` on that, then move the checks in there to throw an exception if `baseUrl` or `forceUrl` are blank. This could also verify that `requestQueue` is non-null.)

sentinel/src/main/java/gov/va/health/api/sentinel/crawler/UrlReplacementRequestQueue.java
```diff
@@ -0,0 +1,41 @@
+package gov.va.health.api.sentinel.crawler;
+
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+import lombok.Builder;
+
```
bschofield-va: Maybe a little javadoc explaining what this does and why you'd use it?

sentinel/src/main/java/gov/va/health/api/sentinel/crawler/UrlReplacementRequestQueue.java
```diff
@@ -0,0 +1,41 @@
...
+public class UrlReplacementRequestQueue implements RequestQueue {
+
+  String baseUrl;
```
bschofield-va: Package level access is too open. These should be `private` and `final` since no one outside of this class has need of them and they will not change after this object is created.

sentinel/src/main/java/gov/va/health/api/sentinel/crawler/UrlReplacementRequestQueue.java
```diff
@@ -0,0 +1,41 @@
...
+  public String next() {
+    return requestQueue.next();
+  }
+
+  @Builder
+  UrlReplacementRequestQueue(String baseUrl, String forceUrl, RequestQueue requestQueue) {
+    this.baseUrl = baseUrl;
+    this.forceUrl = forceUrl;
+    this.requestQueue = requestQueue;
+    if (isBlank(forceUrl)) {
```
bschofield-va: In general, you check arguments _before_ you use them. Move these statements to the top of the constructor.

sentinel/src/main/java/gov/va/health/api/sentinel/crawler/UrlReplacementRequestQueue.java
```diff
@@ -0,0 +1,41 @@
+package gov.va.health.api.sentinel.crawler;
+
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+import lombok.Builder;
+
+public class UrlReplacementRequestQueue implements RequestQueue {
+
+  String baseUrl;
+  String forceUrl;
```
bschofield-va: I challenge the names _baseUrl_ and _forceUrl_. Which is more understandable?

```
        UrlReplacementRequestQueue.builder()
            .forceUrl("https://staging-argonaut.lighthouse.va.gov/api/")
            .baseUrl("https://dev-api.va.gov/services/argonaut/v0/")
            .requestQueue(q)
            .build();
```
```
        UrlReplacementRequestQueue.builder()
            .replaceUrl("https://dev-api.va.gov/services/argonaut/v0/")
            .withUrl("https://staging-argonaut.lighthouse.va.gov/api/")
            .requestQueue(q)
            .build();
```

PR #161, opened 2019-02-11
API-797 add java arrangement to build

jblefkove-va:
Add arranger plugin to build. See corresponding [PR for jarranger-maven-plugin](https://github.com/department-of-veterans-affairs/jarranger-maven-plugin/pull/1).


pom.xml
```diff
@@ -191,6 +191,20 @@
               </execution>
             </executions>
           </plugin>
+          <!-- Java arrangement -->
+          <plugin>
+            <groupId>gov.va.jarranger</groupId>
+            <artifactId>jarranger-maven-plugin</artifactId>
+            <version>0.1.0-SNAPSHOT</version>
```
jblefkove-va: This will be a release version before merging. The plugin has not been officially built by Jenkins yet.

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Practitioner.java
```diff
@@ -58,25 +58,43 @@
...
+  public enum Gender {
+    male,
+    female,
+    other,
+    unknown
+  }
+
   @Data
   @NoArgsConstructor
   @EqualsAndHashCode(callSuper = true)
```
esubocz-va: Wondering how Entry ended up being sorted above Bundle?

jblefkove-va: Only the methods are alphabetized. Other types are grouped together, but the existing order is preserved. (Enums sort before inner classes though, which is why this file got touched.)

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/Location.java
```diff
@@ -55,23 +55,48 @@
   String language;
 
   @Valid Narrative text;
+
```
bschofield-va: Does adding a space between every field fight with the formatter that seems to collapse spaces for many fields?

jblefkove-va: The JavaParser library prints blank lines after fields when it writes the files back to disk. I don't think that is configurable. IntelliJ has rules for blank lines that are applied when it arranges the code; by default it removes blank lines around fields. (The Google formatter generally preserves whatever blank lines you have without adding/removing any. Unfortunately IntelliJ still uses its own rules even if you have the Google formatter plugin.)

You can configure IntelliJ to make the tools match:
`Settings / Code Style / Java / Blank Lines / Minimum Blank Lines / Around field = 1`

jblefkove-va:
The Jarranger plugin has been released, so this is ready for an official review.

bschofield-va:
This change will require use of the VA Nexus repository. If you have not yet configured your local maven settings, you will need to do this now.

Our Nexus server hosted within the VA is now available from outside of the VA network.
https://tools.health.dev-developer.va.gov/nexus

:cherries: Access to this server requires Github.com Personal Access Token with `read:org` permission.

To log in to this server, you need to use your Github.com account associated with the Department of Veterans Affairs org along with the access token, e.g.
```
bschofield-va
3a8618fb338c2fae885131a7e81d5f5f68e8578f
```

:tangerine:To use this new repo with Maven, you must update your ~/.m2/settings.xml file to include `health-apis-releases` and the corresponding `gov.va.api.health` profile.
See https://github.com/department-of-veterans-affairs/health-apis-devops/blob/master/ci/pipeline/resources/maven/settings.xml

:mushroom: If you experience certificate errors, this helper script can be used to update your JDK's cacerts.
https://github.com/department-of-veterans-affairs/health-apis/blob/master/src/scripts/trust-dev-certs.sh

@ilaflamme-va @eclendenning-va @esubocz-va 
@jgarnhart-va @jjennings-va @alecbert-va @dbrown-va 
@dtournour-va

PR #162, opened 2019-02-13
Task/api 741 clinician sentinel tests

eclendenning-va:
Related Stories :
- https://vasdvp.atlassian.net/browse/API-741
- https://vasdvp.atlassian.net/browse/API-746

Tested against Argonaut (Argo)

***Still needs to be tested against Clinician Argonaut (Cargo)***


bschofield-va:
I'll update the upgraderator to support `--include-categories` to match this behavior

PR #165, opened 2019-02-14
Task/API-743 99 Red Patients

esubocz-va:
Allow for patients to be specified as a command line argument. 
Related story: https://vasdvp.atlassian.net/browse/API-743

### DIFF
--- a/sentinel/src/test/java/gov/va/health/api/sentinel/LabCrawlerTest.java
+++ b/sentinel/src/test/java/gov/va/health/api/sentinel/LabCrawlerTest.java
@@ -27,9 +27,17 @@ public class LabCrawlerTest {
 
   @Category({NotInLocal.class, NotInLab.class, NotInProd.class})
   @Test
-  public void crawl() {
+  public void crawlPatients() {
     SystemDefinition env = Sentinel.get().system();
-    String patient = System.getProperty("patient-id", "vasdvp+IDME_01@gmail.com");
+    String patientIds = System.getProperty("patient-id", "vasdvp+IDME_01@gmail.com");
+
+    String[] patients = patientIds.split("\\s*,\\s*");
+    for (String patient : patients) {
+      crawl(env, patient);
+    }
+  }
+
+  private void crawl(SystemDefinition env, String patient) {
     UserCredentials user =
         UserCredentials.builder()
             .id(patient)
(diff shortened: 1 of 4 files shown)


sentinel/src/test/java/gov/va/health/api/sentinel/LabCrawlerTest.java
```diff
@@ -47,36 +58,33 @@ private void crawl(IdMeOauthRobot robot) {
...
   }
 
   @Category(Manual.class)
   @Test
-  public void crawlUser1() {
-    crawl(robots.user1());
+  public void crawlPatients() {
+    SystemDefinition env = Sentinel.get().system();
+    String patientIds = System.getProperty("patient-id", "vasdvp+IDME_01@gmail.com");
+    String[] patients = patientIds.split("\\s*,\\s*");
```
jblefkove-va: This regex won't trim whitespace from the beginning of the first patient or the end of the last one. It may be simpler to do `split(",")` and then call `trim()` on each patient ID before using them.

sentinel/src/test/java/gov/va/health/api/sentinel/LabCrawlerTest.java
```diff
@@ -47,36 +58,33 @@ private void crawl(IdMeOauthRobot robot) {
...
+    for (String patient : patients) {
+      crawl(env, patient);
+    }
+    assertThat(failureCount).withFailMessage("%d Failures", failureCount).isEqualTo(0);
```
jblefkove-va: This test would be a little easier to understand if `failureCount` were a local variable here. Instead of having `crawl(env, patient)` add to `failureCount` at the end, return the number of failures and sum them in the above for loop.

sentinel/src/test/java/gov/va/health/api/sentinel/LabCrawlerTest.java
```diff
@@ -47,36 +58,33 @@ private void crawl(IdMeOauthRobot robot) {
...
-    assertThat(results.failures()).withFailMessage("%d Failures", results.failures()).isEqualTo(0);
+    failureCount += results.failures();
   }
 
   @Category(Manual.class)
   @Test
-  public void crawlUser1() {
-    crawl(robots.user1());
+  public void crawlPatients() {
+    SystemDefinition env = Sentinel.get().system();
```
jblefkove-va: Is there an advantage to defining `env` once here and passing it into each invocation of `crawl(env, patient)`? (Vs. calling `Sentinel.get().system()` each time in the inner crawl method.)

PR #166, opened 2019-02-15
API-814 procedure hack jargonaut

jblefkove-va:
https://vasdvp.atlassian.net/browse/API-814


argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/procedure/ProcedureController.java
```diff
@@ -44,9 +53,24 @@
 )
 @AllArgsConstructor(onConstructor = @__({@Autowired}))
 public class ProcedureController {
+  /**
+   * Optional patient ID with procedure data that can secretly service requests for {@link
+   * #superman}.
+   */
+  @Value("${clark-kent:}")
```
bschofield-va: I think we need to make these properties a little more clear since they will be maintained in properties files independently. Perhaps something like this?

```
procedure.test-patient-workaround.id-without-records=12345v67890
procedure.test-patient-workaround.id-with-records=112233v445566
```

PR #167, opened 2019-02-15
Task/api 865 lab users

bschofield-va:
Adds a manual test that can be used to validate which lab users can log in.
Adds support for OAuth flow to support "allow access" forms and explicitly look for error pages.

### DIFF
--- a/sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
+++ b/sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
@@ -98,7 +98,7 @@ public class IdMeOauthRobot {
     url = waitForUrlToChange(driver, url);
     // Continue passed entering the authentication code
     driver.findElement(By.className("btn-primary")).click();
-    url = waitForUrlToChange(driver, url);
+    waitForUrlToChange(driver, url);
   }
 
   /** Return the authorization code, logging in if necessary. */
@@ -216,17 +216,13 @@ public class IdMeOauthRobot {
 
   /** Waits for the current page to completely load. */
   public void waitForPageLoad(WebDriver driver) {
-    ExpectedCondition<Boolean> pageLoadCondition =
-        new ExpectedCondition<Boolean>() {
-          @Override
-          public Boolean apply(WebDriver driver) {
-            return ((JavascriptExecutor) driver)
-                .executeScript("return document.readyState")
-                .equals("complete");
-          }
-        };
     WebDriverWait wait = new WebDriverWait(driver, 30);
-    wait.until(pageLoadCondition);
+    wait.until(
+        (ExpectedCondition<Boolean>)
+            d ->
+                ((JavascriptExecutor) driver)
+                    .executeScript("return document.readyState")
+                    .equals("complete"));
   }
 
   private String waitForUrlToChange(WebDriver driver, String url) {
--- a/sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
+++ b/sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
@@ -164,10 +164,9 @@ public class IdMeOauthRobot {
   }
 
   private String extractCodeFromRedirectUrl(WebDriver driver) {
-    String url;
     new WebDriverWait(driver, 1, 100)
         .until(ExpectedConditions.urlContains(config.authorization().redirectUrl()));
-    url = driver.getCurrentUrl();
+    String url = driver.getCurrentUrl();
     log.info("Redirected {}", url);
 
     return Arrays.stream(url.split("\\?")[1].split("&"))
@@ -215,7 +214,7 @@ public class IdMeOauthRobot {
   }
 
   /** Waits for the current page to completely load. */
-  public void waitForPageLoad(WebDriver driver) {
+  private void waitForPageLoad(WebDriver driver) {
     WebDriverWait wait = new WebDriverWait(driver, 30);
     wait.until(
         (ExpectedCondition<Boolean>)
--- a/sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
+++ b/sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
@@ -101,6 +101,7 @@ public class IdMeOauthRobot {
     url = waitForUrlToChange(driver, url);
   }
 
+  /** Return the authorization code, logging in if necessary. */
   @SneakyThrows
   public String code() {
     if (code != null) {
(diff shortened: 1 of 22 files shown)


sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
```diff
@@ -21,32 +24,133 @@
...
+    driver.findElement(By.className("btn-primary")).click();
+    url = waitForUrlToChange(driver, url);
+    // Continue passed entering the authentication code
+    driver.findElement(By.className("btn-primary")).click();
+    waitForUrlToChange(driver, url);
+  }
+
+  /** Return the authorization code, logging in if necessary. */
+  @SneakyThrows
+  public String code() {
```
jblefkove-va: Do this method and `waitForPageLoad()` need to be public? The only one I see used externally is `token()`.

sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
```diff
@@ -21,32 +24,133 @@
...
+
+      String url = driver.getCurrentUrl();
+      enterCredentials(driver);
+      checkForBadCredentials(driver);
+      url = waitForUrlToChange(driver, url);
+      clickThroughFakeTwoFactorAuthentication(driver);
+      /*
+       * There are possibly two consent forms.
+       */
+      url = waitForUrlToChange(driver, url);
```
jblefkove-va: There is also a call to `waitForUrlToChange()` at the end of `clickThroughFakeTwoFactorAuthentication()`. Do you need to do it twice?

sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
```diff
@@ -57,33 +161,38 @@ private String authorize() {
...
-    log.info("Clicking through two factor authorization sham");
-    driver.findElement(By.className("btn-primary")).click();
-    // Continue passed entering the authentication code
-    driver.findElement(By.className("btn-primary")).click();
-
-    String url = driver.getCurrentUrl();
+  }
 
+  private String extractCodeFromRedirectUrl(WebDriver driver) {
+    String url;
```
jblefkove-va: Does `url` need to be defined up front?

sentinel/src/test/java/gov/va/health/api/sentinel/LabLogInTest.java
```diff
@@ -0,0 +1,101 @@
...
+      ids.add("va.api.user+idme." + String.format("%03d", i) + "@gmail.com");
+    }
+
+    List<String> winners = new CopyOnWriteArrayList<>();
+    List<String> losers = new CopyOnWriteArrayList<>();
+    LabRobots robots = LabRobots.fromSystemProperties();
+
+    ExecutorService ex = Executors.newFixedThreadPool(10);
+
+    List<Future<?>> futures = new LinkedList<>();
```
jblefkove-va: This doesn't need to be a `LinkedList`, you know how many elements are going in here 😛. For that matter `ids` doesn't need to be a linked list either.

sentinel/src/test/java/gov/va/health/api/sentinel/LabLogInTest.java
```diff
@@ -0,0 +1,101 @@
...
+                } catch (Exception e) {
+                  log.error(
+                      "LOSER: {} ({}/{})", e.getMessage(), id, robots.labConfig().userPassword());
+                  losers.add(id + " - " + e.getMessage());
+                }
+              }));
+    }
+
+    ex.shutdown();
+    ex.awaitTermination(10, TimeUnit.MINUTES);
```
jblefkove-va: I don't know all the intricacies of Java multi-threading; is there an advantage to shutting down the executor service before calling `get()` on all of the futures? Is there an advantage to explicitly shutting down the executor service vs. letting it die when the test ends?

sentinel/src/test/java/gov/va/health/api/sentinel/LabLogInTest.java
```diff
@@ -0,0 +1,101 @@
...
+          } catch (Exception e) {
+            log.error(e.getMessage());
+          }
+        });
+
+    String report =
+        Stream.concat(winners.stream().map(w -> w + " - OK"), losers.stream())
+            .sorted()
+            .collect(Collectors.joining("\n"));
+    Files.write(report.getBytes(StandardCharsets.UTF_8), new File("lab-users.txt"));
```
jblefkove-va: Did you mean to use the Guava `Files` here instead of Java native `Files`? (I don't have a preference, but I don't see Guava used anywhere else in the code.)

sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
bschofield-va: yes

sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
bschofield-va: no

sentinel/src/test/java/gov/va/health/api/sentinel/LabLogInTest.java
bschofield-va: I fixed futures, but I'll leave `ids`.  While debugging logins, it's very helpful to modify loop values... and performance implications are minimal in this context

sentinel/src/test/java/gov/va/health/api/sentinel/LabLogInTest.java
bschofield-va: good catch

sentinel/src/test/java/gov/va/health/api/sentinel/LabLogInTest.java
bschofield-va: No advantage between shutting down and `get`. 

1. `shutdown()` does not to stop pending tasks, they will be allowed to continue (but it prevent new tasks from being added).  The `shutdown()` call is so that `awaitTermination(...)` can be used.
2. `awaitTermination(...)` causes an exception to occur if the login tasks take more than 10 minutes
3. The future `get()` is to flush out any exceptions that might have been thrown by individual tasks. Exceptions thrown by tasks are suppressed by the ExecutorService until `get()` is called.

Since login tasks are occurring in background threads, I need to wait for all of them to complete before checking for failures. Alternatively to the `shutdown+awaitTermination` calls, I could have put time limits on `future.get(60,TimeUnit.SECONDS)` to make sure we don't wait forever.

sentinel/src/main/java/gov/va/health/api/sentinel/IdMeOauthRobot.java
bschofield-va: I'm going to leave `code` public since it is relevant information from an OAuth perspective and can be useful for future robots

`waitForPageLoad` is a mistake

PR #172, opened 2019-02-21
API-873 Add time limit to crawler

jblefkove-va:
https://vasdvp.atlassian.net/browse/API-873

Add configurable time limit to crawler. Change crawler request queue to prioritize resources that have been visited less frequently. `SummarizingResultCollector` prints a summary of resource reads and searches.


sentinel/src/main/java/gov/va/health/api/sentinel/crawler/ResourceDiscovery.java
```diff
@@ -40,10 +43,41 @@ private ResourceDiscovery(@NonNull String url, @NonNull String patientId) {
...
+   * <p>For example, resource is 'Condition' in these URLs:
+   *
+   * <ul>
+   *   <li>foo/api/Condition
+   *   <li>foo/api/Condition/123
+   *   <li>foo/api/Condition?patient=bobnelson
+   * </ul>
+   */
+  static String resource(@NonNull String url) {
+    Matcher matcher = Pattern.compile(".*/api/([^/^?]+).*").matcher(url);
```
eclendenning-va: `/api/` is not the apiPath defined for the lab environment so this will cause the the lab crawler test to warn about failing to extract the resource, but still pass when it shouldn't.

jblefkove-va: I updated the `resource()` method not to depend on `api` in the URL. It is still parsing out the resource without any context about what the URL base path is, which is tricky.

It will work for URLs of these two formats:

- `https://foo.gov/services/argonaut/v0/Condition?patient=12345`
- `https://foo.gov/api/Condition/12345`

Do we expect to see URLs of any other formats, e.g. `https://foo.gov/services/argonaut/v0/Condition/12345?patient=12345` ?

bschofield-va: For reads like `https://foo.gov/services/argonaut/v0/Condition/12345`, the spec does indicate that some query parameters are possible, but we don't support any at this time, e.g. `https://foo.gov/services/argonaut/v0/Condition/12345?_summary=text`

http://hl7.org/fhir/DSTU2/http.html#read

sentinel/src/main/java/gov/va/health/api/sentinel/crawler/Crawler.java
```diff
@@ -69,20 +74,31 @@ private String asAdditionalInfo(Exception e) {
...
   public void crawl() {
+    Stopwatch watch = Stopwatch.createStarted();
     results.init();
-    List<Future<?>> futures = new LinkedList<>();
-
+    Stack<Future<?>> futures = new Stack<>();
     ScheduledExecutorService monitor = monitorPendingRequests(futures);
 
     while (hasPendingRequests(futures)) {
+      if (timeLimit != null && watch.elapsed(TimeUnit.SECONDS) > timeLimit.getSeconds()) {
```
bschofield-va: Would it possible to combine this into one clause?

```
while( hasPendingRequests(futures) && hasNotExceededTimeLimit(futures) ) {
   ... process next request ...
}
```

PR #173, opened 2019-02-21
Task/api 804 well known

jjennings-va:
https://vasdvp.atlassian.net/browse/API-804

Added in .well-known endpoint that will provide the following required fields
- Authorization endpoint
- Token endpoint
- Capabilities (Please look over the capabilities here ( http://hl7.org/fhir/smart-app-launch/conformance/index.html#using-well-known) 

The following optional fields are also included in this PR

- responseTypeSupported
- scopesSupported


argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/WellKnown.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@Schema(description = "Well-Known")
```
dbrown-va: http://www.hl7.org/fhir/smart-app-launch/conformance/#using-well-known

Probably worth while to call out the FHIR reference.

argonaut/src/test/java/gov/va/api/health/argonaut/service/controller/wellknown/WellKnownControllerTest.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  private WellKnownProperties properties() {
+    return WellKnownProperties.builder()
+        .security(
+            WellKnownProperties.SecurityProperties.builder()
+                .tokenEndpoint("https://argonaut.lighthouse.va.gov/token")
+                .authorizeEndpoint("https://argonaut.lighthouse.va.gov/authorize")
+                .build())
+        .capabilities("context-standalone-patient launch-ehr permission-offline permission-patient")
```
alecbert-va: As a test this is fine but in the real deal I think launch-ehr is actually going to be launch-patient

argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/wellknown/WellKnownController.java
```diff
@@ -0,0 +1,32 @@
...
+class WellKnownController {
+
+  private final WellKnownProperties properties;
+
+  @GetMapping
+  WellKnown read() {
+    return WellKnown.builder()
+        .authorizationEndpoint(properties.getSecurity().getAuthorizeEndpoint())
+        .tokenEndpoint(properties.getSecurity().getTokenEndpoint())
+        .capabilities(asList(properties.getCapabilities().split(" ")))
```
bschofield-va: All of these `.split(" ")` calls are potential NPEs since none of those fields are guaranteed non-null. Consider using a small helper method to handle the splits... Also consider using csv to split since that is more common than space delimited.

```
List<String> split(String csv) {
  // deal with null
  // split and deal with extra whitespacings
}

```

argonaut/src/test/java/gov/va/api/health/argonaut/service/controller/wellknown/WellKnownControllerTest.java
jjennings-va: went ahead and renamed since I was in there anyway

argonaut-api/src/main/java/gov/va/api/health/argonaut/api/resources/WellKnown.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
+@Schema(description = "http://hl7.org/fhir/smart-app-launch/conformance/index.html#using-well-known")
+public final class WellKnown {
```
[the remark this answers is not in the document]
jjennings-va: For now, I packaged it under "information". I think this makes a little more sense then leaving it inside of resource. If you have a better location for it, I am willing to move it again.

argonaut/src/main/resources/application.properties
```diff
@@ -18,6 +18,11 @@ management.endpoints.web.exposure.include=health,info,env,metrics,prometheus
 management.endpoint.health.show-details=ALWAYS
 management.health.diskspace.enabled=false
 spring.jackson.default-property-inclusion=non_empty
+well-known.security.authorize-endpoint=unset
```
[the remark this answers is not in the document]
jjennings-va: So... I agree that duplication of the properties was a bad idea. I decided to go with renaming the new properties I created for well-known to use "conformance" as the prefix within the properties instead of "well-known". This will eliminate any duplication and makes sense to me as it will not require changing any other property names.

argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/wellknown/WellKnownController.java
bschofield-va: Actually this is not necessary at all... Spring supports lists in properties, so you need something like this:

```
class WellKnownProperties {
  List<String> capabilities;
}
```
```
wellknown.capabilities = context-standalone-patient, launch-ehr, permission-offline, permission-patient
```

argonaut/src/test/java/gov/va/api/health/argonaut/service/controller/wellknown/WellKnownControllerTest.java
```diff
@@ -0,0 +1,46 @@
...
+        .scopesSupported("patient/DiagnosticReport.read patient/Patient.read offline_access")
+        .build();
+  }
+
+  @Test
+  @SneakyThrows
+  public void read() {
+    WellKnownProperties properties = properties();
+    WellKnownController controller = new WellKnownController(properties);
+    WellKnown old =
```
bschofield-va: Instead of loading a file to create you expected value, you should just create it using builders and reduce the number of moving parts by one file.

jgarnhart-va:
@bschofield-va - Looks like all feedback is addressed. One more check?

argonaut/src/main/resources/application.properties
```diff
@@ -18,6 +18,9 @@ management.endpoints.web.exposure.include=health,info,env,metrics,prometheus
 management.endpoint.health.show-details=ALWAYS
 management.health.diskspace.enabled=false
 spring.jackson.default-property-inclusion=non_empty
+conformance.capabilities=unset
+conformance.responseTypeSupported=unset
+conformance.scopesSupported=unset
```
jblefkove-va: The `make-configs.sh` script should be updated with these new properties.

argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/wellknown/WellKnownProperties.java
```diff
@@ -0,0 +1,37 @@
...
+  private SecurityProperties security;
+  private List<String> responseTypeSupported;
+  private List<String> scopesSupported;
+
+  @Data
+  @Accessors(fluent = false)
+  @NoArgsConstructor
+  @AllArgsConstructor
+  @Builder
+  public static class SecurityProperties {
```
bschofield-va: Leverage the existing ConformanceStatement object instead of duplicating these properties.

argonaut/src/main/resources/application.properties
```diff
@@ -18,6 +18,9 @@ management.endpoints.web.exposure.include=health,info,env,metrics,prometheus
 management.endpoint.health.show-details=ALWAYS
 management.health.diskspace.enabled=false
 spring.jackson.default-property-inclusion=non_empty
+conformance.capabilities=unset
```
bschofield-va: these should be `well-known` and prefer kabab case for properties, e.g. `response-type-supported`

sentinel/src/test/java/gov/va/health/api/sentinel/WellKnownIT.java
```diff
@@ -0,0 +1,26 @@
...
+  @Category({Local.class, LabArgo.class, LabCargo.class, ProdArgo.class, ProdCargo.class})
+  public void wellKnownIsValid() {
+    ExpectedResponse response =
+        Sentinel.get().clients().argonaut().get(".well-known/smart-configuration");
+    response.expect(200);
+    String rawJson = response.response().asString();
+    assertThat(rawJson)
+        .withFailMessage("Tabs and newlines break our customer.")
+        .doesNotContain("\n", "\t");
+  }
```
bschofield-va: You could compare auth and token endpoints with conformance statement to make sure they are in sync.

argonaut/src/main/java/gov/va/api/health/argonaut/service/controller/wellknown/WellKnownProperties.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.NoArgsConstructor;
+import lombok.experimental.Accessors;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@SuppressWarnings("DefaultAnnotationParam")
+@Configuration
+@EnableConfigurationProperties
+@ConfigurationProperties("conformance")
```
bschofield-va: `well-known` instead of `conformance`

jjennings-va:
@jblefkove-va @bschofield-va 

Addressed feedback, please let me know if there are any other changes or issues you would like me to make.

PR #175, opened 2019-02-25
Task/API-882 Edit url replacement request queue builder to be agnostic about /s

esubocz-va:
Make /'s agnostic in sentinel crawlers to prevent missing or double /'s in url replacement.


jblefkove-va:
Let's add a unit test for this.

PR #184, opened 2019-03-26
API-998 database refactoring POC for diagnostic reports

jblefkove-va:
https://vasdvp.atlassian.net/browse/API-998

Proof of concept for consuming diagnostic reports using the new simplified database schema that avoids the `prc_resource_return` stored procedure.

Notes:
* `data-query` now uses `ids` directly. The code for converting between public IDs and CDW IDs has been copied from `mr-anderson` into `data-query`. 
   * Rather than refactor, this duplication is intended to be temporary, as `mr-anderson` will be decommissioned once all resources are switched to the new approach.
* This is intended to run alongside the existing behavior in shadow mode, based on the value of the `Database-2-0-Mode` header in the request.
  * If the header is `true`, the new approach is used.
  * If the header is `both`, both the old and new ways are executed. A warning is logged if the results differ. The results from the old approach are still returned.
  * Otherwise, the current approach is used with no change in behavior.
* Code coverage temporarily reduced. No new unit tests yet.
* Contact Mark Shaffer for access to a database instance with the new schema.


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
```diff
@@ -106,53 +108,54 @@ private void addQueryParameters(TypedQuery<?> query) {
 
   private Instant computeLowerBound(String paramString) {
     log.error("Computing lower bound for " + paramString);
```
jgarnhart-va: Why is this an error?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
```diff
@@ -106,53 +108,54 @@ private void addQueryParameters(TypedQuery<?> query) {
...
       case 25:
-        // .dateYearMonthDayHourMinuteSecondTimezone("1970-01-01T07:00:00+05:00")
+        return OffsetDateTime.parse(paramString).toInstant();
       default:
         throw new IllegalArgumentException();
     }
   }
 
   private Instant computeUpperBound(String paramString) {
     log.error("Computing upper bound for " + paramString);
```
jgarnhart-va: Why error?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
```diff
@@ -106,53 +108,54 @@ private void addQueryParameters(TypedQuery<?> query) {
...
+    OffsetDateTime theLowerBound =
+        OffsetDateTime.ofInstant(computeLowerBound(paramString), ZonedDateTime.now().getOffset());
     switch (paramString.length()) {
       case 4:
```
jgarnhart-va: This case statement might read better if we had constants representing these numbers. So for case 4, it looks like that could be something like "YEAR". Case 7 could be "YEAR_MONTH", etc.

data-query/pom.xml
```diff
@@ -20,6 +33,11 @@
       <artifactId>data-query-api</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>gov.va.api.health</groupId>
+      <artifactId>ids-client</artifactId>
+      <version>1.1.0</version>
+    </dependency>
     <dependency>
       <groupId>gov.va.dvp</groupId>
```
jgarnhart-va: I think as we move away from obtaining XML from the  DB, this dependency will no longer be needed. Just something to keep in mind.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
jblefkove-va: Sounds good, I will update.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
jblefkove-va: It's not really an error; these are temporary log statements for development. They're at `error` level so I'm less likely to forget about them... I will remove or downgrade to `info` before resolving the PR.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
```diff
@@ -0,0 +1,234 @@
...
+    return "date" + index + "UpperBound";
+  }
+
+  private String toQuerySnippet() {
+    switch (prefix()) {
+      case EQ:
+        // the range of the search value fully contains the range of the target value
+        return String.format(
+            " and :%s <= dr.effectiveDateTime and dr.issuedDateTime <= :%s",
+            lowerBoundPlaceholder(), upperBoundPlaceholder());
```
jblefkove-va: Here's something interesting: Spotbugs was warning that `DiagnosticReportController` was vulnerable to JPQL injection. At first I was confused, because I only use `entity.setParameter()` for all of the search params, which is safe. It was actually complaining about these lines here; the query snippet is directly built up using the the two `placeholder()` methods, which return strings like `date0LowerBound`. Previously it was doing a concatenation of a `name` field with a constant `LowerBound` or `UpperBound`, where `name` was provided by `DiagnosticReportController`. `name` would only ever be `date0` or `date1`, so this was still a false positive, but I'm really impressed that SpotBugs detected it at all. (I resolved the warning by changing this class to use an `index` field instead of `name`.)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
```diff
@@ -0,0 +1,247 @@
...
+          .addQueryParameters(query);
+    }
+  }
+
+  private static UnsupportedOperationException apException() {
+    return new UnsupportedOperationException("AP search prefix not implemented");
+  }
+
+  /** Build a combined JPA query snippet representing all the date criteria. */
+  public static String querySnippet(String[] dates) {
```
bschofield-va: What happens if I give three dates? I we need some protection.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
```diff
@@ -0,0 +1,247 @@
...
+
+  SearchPrefix prefix;
+
+  String date;
+
+  @Builder
+  private JpaDateTimeParameter(int index, String paramString) {
+    super();
+    this.index = index;
+    if (Character.isLetter(paramString.charAt(0))) {
```
bschofield-va: We do have parameter validation happening at the rest level, but should we at least check that the paramString has length > 2?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
jblefkove-va: I updated this method and `addQueryParameters` to handle arbitrarily many dates. (The API validation does limit you to two, though.)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/JpaDateTimeParameter.java
jblefkove-va: Added `IllegalArgumentException` if the paramString is too short.

jblefkove-va:
Implementation of date search is complete. The API validation allows the following searches:

* year
* year,month
* year,month,day
* year,month,day,hour,minute,second,timezone/offset

Originally I thought other partial searches (e.g. hour only) were also supported, and were implemented in the `lowerBound()` and `upperBound()` methods in `JpaDateTimeParameters`. I have removed those now.

PR #185, opened 2019-04-11
Bug/api 1050 jargo lab conformance statement

eclendenning-va:
Related Story: https://vasdvp.atlassian.net/browse/API-1050

Used the properties for removing disabled resource references to disable the resources in the conformance statement.


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/conformance/MetadataControllerTest.java
```diff
@@ -51,12 +46,23 @@ private ConformanceStatementProperties properties() {
...
   @Test
   @SneakyThrows
   public void readClinician() {
-    ConformanceStatementProperties properties = properties();
-    properties.setStatementType(StatementType.CLINICIAN);
-    MetadataController controller = new MetadataController(properties);
+    ConformanceStatementProperties conformanceStatementProperties =
+        conformanceStatementProperties();
+    ReferenceSerializerProperties referenceSerializerProperties =
+        referenceSerializerPropertiesTrue();
```
jblefkove-va: How about one `referenceSerializerProperties()` method that accepts a boolean parameter?

PR #188, opened 2019-04-29
ADQ-5-upgrade-data-querys-health-actuator

pscorer:
In addition to ADQ-5, fixed minor issue with developer script dev-app.sh.


.gitignore
```diff
@@ -27,3 +27,4 @@
 target/
 
 secrets.conf
+*.pem
```
jblefkove-va: Is this for the certificates from `trust-dev-certs.sh` ?

jblefkove-va:
For future stories, even though you are forking, you should still use a branch instead of committing to master.

.gitignore
pscorer: Yes. OK to make change?

jblefkove-va: :shipit:

PR #189, opened 2019-04-30
Task/ADQ-1 Signature datatype cardinality

CameronPH:
JIRA Link: https://vasdvp.atlassian.net/browse/ADQ-1

Changed the validation annotation for the fields `whoUri` and `whoReference` of Signature.java from `@ZeroOrOneOf` to `@ExactlyOneOf` to correct the cardinality of the datatype.

Added unit test `SignatureExactlyOneOfTest.java` which verifies the fix.


data-query-api/src/test/java/gov/va/api/health/dataquery/api/SignatureExactlyOneOfTest.java
```diff
@@ -0,0 +1,39 @@
...
+	public void test() {		
+		Set<ConstraintViolation<Signature>> problems = Validation
+				.buildDefaultValidatorFactory()
+				.getValidator()
+				.validate(psuedoSignature);
+		
+		assertEquals(problems.size(), 0);
+	}
+	
+//	Sample Signature
```
jblefkove-va: This comment is superfluous and can be removed. If you do want to put a comment on a method, use `/** javadoc */`.

data-query-api/src/test/java/gov/va/api/health/dataquery/api/SignatureExactlyOneOfTest.java
```diff
@@ -0,0 +1,39 @@
...
+	private Signature signature() {
+	    return Signature.builder()
+	        .id("0714")
+	        .extension(singletonList(data.extension()))
```
jblefkove-va: It's valid to have a `Signature` with no extensions. This doesn't hurt anything, but it does add unnecessary complexity to the test; I would remove it.

data-query-api/src/test/java/gov/va/api/health/dataquery/api/SignatureExactlyOneOfTest.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Set;
+import javax.validation.Validation;
+import javax.validation.ConstraintViolation;
+import org.junit.Test;
+import gov.va.api.health.dataquery.api.datatypes.Signature;
+import gov.va.api.health.dataquery.api.samples.SampleDataTypes;
+
+public class SignatureExactlyOneOfTest {
+	private final SampleDataTypes data = SampleDataTypes.get();
+	private final Signature psuedoSignature = signature();
```
jblefkove-va: Is there a reason for these class fields vs. using local variables in the test method? (In general, the more you can localize everything in the test method itself, the easier the test is to read and understand.)

data-query-api/src/test/java/gov/va/api/health/dataquery/api/SignatureExactlyOneOfTest.java
```diff
@@ -0,0 +1,39 @@
...
+	
+	@Test
+	public void test() {
```
jblefkove-va: The test method name should have some description of what is being validated. In this case, something like `signatureWithOnlyWhoReferenceIsValid`

data-query-api/src/test/java/gov/va/api/health/dataquery/api/SignatureExactlyOneOfTest.java
```diff
@@ -0,0 +1,39 @@
...
+	private final Signature psuedoSignature = signature();
+	
+	@Test
+	public void test() {		
+		Set<ConstraintViolation<Signature>> problems = Validation
+				.buildDefaultValidatorFactory()
+				.getValidator()
+				.validate(psuedoSignature);
+		
+		assertEquals(problems.size(), 0);
```
jblefkove-va: Use `org.assertj.core.api.Assertions.assertThat`. Look at some of the other tests for examples. (This is more readable. One of the downsides of the JUnit assert is that the order of argument is expected, then actual. This is backwards from how most people think about; when the test is passing, the order doesn't matter, but if the test fails, the error message will be confusing, because the parameters are backwards.)

data-query-api/src/test/java/gov/va/api/health/dataquery/api/SignatureExactlyOneOfTest.java
```diff
@@ -0,0 +1,39 @@
...
+	}
+	
+//	Sample Signature
+	private Signature signature() {
+	    return Signature.builder()
+	        .id("0714")
+	        .extension(singletonList(data.extension()))
+	        .type(data.codingList())
+	    	.when("2000-01-01T00:00:00-00:00")
+	    	.whoUri(null)
```
jblefkove-va: You don't need to explicitly set this to `null`, any parameters you don't explicitly set in the builder will default to `null`.

data-query-api/src/test/java/gov/va/api/health/dataquery/api/datatypes/SignatureTest.java
```diff
@@ -0,0 +1,110 @@
...
+    Set<ConstraintViolation<Signature>> problems = Validation
+        .buildDefaultValidatorFactory()
+        .getValidator()
+        .validate(psuedoSignature);
+
+    assertThat(problems.size()).isEqualTo(1);
+  }
+
+  private Signature signatureWithOnlyWhoReference() {
+    SampleDataTypes data = SampleDataTypes.get();
```
jblefkove-va: If you have test data that is only used in one place, it's usually cleaner to inline it in the test method.

bschofield-va:
@esubocz-va This might affect you depending on when you branched

esubocz-va:
@alecbert-va

PR #193, opened 2019-05-06
Bug/ADQ-12 fhir.base64regex fails good strings

pscorer:


bschofield-va:
@esubocz-va FYI this affects you

esubocz-va:
The data-query refactoring task has been moved away from me, so @alecbert-va you should probably have a look at this. Thanks for the heads up @bschofield-va

PR #196, opened 2019-05-15
ADQ-22 simplify int and crawler test exec

pscorer:


data-query-tests/src/main/docker/entrypoint.sh
```diff
@@ -4,28 +4,37 @@
...
+    --exclude-category gov.va.api.health.sentinel.categories.Local \
+    --include-category gov.va.api.health.sentinel.categories.Manual \
+    --trust example.us-gov-west-1.elb.amazonaws.com
     -Dlab.client-id=12345 \
     -Dlab.client-secret=ABCDEF \
     -Dlab.user-password=secret \
-    gov.va.api.health.sentinel.LabCrawlerTest
+    gov.va.api.health.sentinel.UsingMagicPatientCrawlerTest
 
+Docker Run Examples
```
jhulbert-va: These examples would also need to include the --env K8S_LOAD_BALANCER=example.com and --env K8S_ENVIRONMENT=qa

data-query-tests/src/test/java/gov/va/api/health/sentinel/WellKnownIT.java
```diff
@@ -14,7 +13,7 @@
   @Test
   @Category({
     Local.class,
-    LabDataQueryPatient.class,
+    // TODO Route is not present in the lab so we can't run it there -LabDataQueryPatient.class,
```
jhulbert-va: Is this TODO meant to be for future work?

data-query-tests/src/main/docker/entrypoint.sh
pscorer: Agreed. Corrected 👍

data-query-tests/src/test/java/gov/va/api/health/sentinel/WellKnownIT.java
pscorer: Updated comment with ADQ number of story that will put this in.

data-query-tests/src/main/java/gov/va/api/health/sentinel/crawler/IgnoreFilterResultCollector.java
```diff
@@ -0,0 +1,94 @@
...
+
+  private final Set<String> ignoredFailureSummaries = new ConcurrentSkipListSet<>();
+
+  private final List<String> failuresToIgnore = new ArrayList<>();
+
+  @Override
+  public void add(Result result) {
+    if (result.outcome() != Outcome.OK) {
+      // If this failure is in one or more of the ignore filters then ignore it instead of failing.
+      if (failuresToIgnore.stream().filter(s -> result.query().endsWith(s)).count() > 0) {
```
bschofield-va: `.count() > 0` will iterate over all filters. Consider `.findFirst().isPresent()` which will stop iterating at the first match.

data-query-tests/src/test/java/gov/va/api/health/sentinel/crawler/IgnoreFilterResultCollectorTest.java
```diff
@@ -0,0 +1,68 @@
...
+    Result badnessResult =
+        Result.builder()
+            .query(
+                "https://somepath.va.gov/services/argonaut/v0/Resource/z8z848z3-35zz-5zz-93zz-z4z8731z1z11")
+            .outcome(Outcome.INVALID_PAYLOAD)
+            .timestamp(Instant.ofEpochMilli(158994000000L))
+            .build();
+    results.add(badnessResult);
+
+    assertThat(results.failures()).isOne();
```
bschofield-va: Add me! `assertThat(results.ignoredFailures()).isZero();`

data-query-tests/src/main/java/gov/va/api/health/sentinel/crawler/IgnoreFilterResultCollector.java
pscorer: Yeah, it did seem a bit awkward and I wasn't loving this so thanks for the feedback. Turned it into an anyMatch() which seemed an even better choice for the if() but can use findFirst() if you still lean that way.

data-query-tests/src/test/java/gov/va/api/health/sentinel/crawler/IgnoreFilterResultCollectorTest.java
pscorer: 👍

PR #197, opened 2019-05-15
do not reset SKIP_CRAWLER to blank on start

pscorer:


bschofield-va:
Change being refactored in another PR

PR #200, opened 2019-05-17
ADQ-34 Repackage health api parent sentinel classes

pscorer:
Refactoring
1. Renamed the package gov.va.api.health.sentinel to gov.api.health.dataquery.tests. With this change there aren't two repos contributing to the same name space.
2. Moved SmokeTest category to the real sentinel
3. Moved FhirTestClient to the real sentinel
4. Had to refactor FhirTestClient to use a method for transient data equality comparison of error response body
5. Moved the crawler properties (replacement url and crawler filter) to the existing CrawlerProperties

This PR has to be merged after the ADQ-34 change in health-apis-parent and will need a downstream change to the [environment].testvars files in the dq deployment unit (Josh has this ready) to account for the changes in namespace for the test classes.


data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/LabRobots.java
```diff
@@ -198,7 +198,8 @@ private String valueOf(String name) {
     String token;
     String authorize;
 
-    SmartOnFhirUrls(String baseUrl) {
+    /** . */
```
bschofield-va: lol. We'll fix this comment in a subsequent PR that does not involve a fork.

PR #204, opened 2019-05-23
Task/adq 40 remove referencerange observation

jhulbert-va:
[ADQ-40](https://vasdvp.atlassian.net/browse/ADQ-40)

* Reference Range is removed if category is 'vital-signs'
* Added Junit to test functionality 

**The db patient is only returning a case with vital-signs so the other case has not been tested. Except by the junit tests.**

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationTransformer.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationTransformer.java
@@ -295,7 +295,7 @@ public class ObservationTransformer implements ObservationController.Transformer
       CdwReferenceRanges maybeCdw, CdwCategory cdwCategory) {
     List<CdwCategory.CdwCoding> categoryCoding = cdwCategory.getCoding();
     if (categoryCoding.size() > 0
-        && !StringUtils.isEmpty(categoryCoding.get(0).getCode().value())
+        && (categoryCoding.get(0).getCode() != null)
         && categoryCoding
             .get(0)
             .getCode()
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationTransformer.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationTransformer.java
@@ -49,6 +49,7 @@ public class ObservationTransformer implements ObservationController.Transformer
      * Specimen reference is omitted since we do not support the a specimen resource and
      * do not want dead links
      */
+
     return Observation.builder()
         .resourceType("Observation")
         .id(cdw.getCdwId())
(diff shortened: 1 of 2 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationTransformer.java
```diff
@@ -64,7 +65,10 @@ public Observation apply(CdwObservation cdw) {
         .valueCodeableConcept(valueCodeableConcept(cdw.getValueCodeableConcept()))
         .interpretation(interpretation(cdw.getInterpretation()))
         .comments(cdw.getComments())
-        .referenceRange(referenceRanges(cdw.getReferenceRanges()))
+        .referenceRange(
+            (cdw.getCategory().getCoding().get(0).getCode().value().equals(CdwObservationCategoryCode.VITAL_SIGNS.value()))
```
eclendenning-va: Possible NPE. Consider extracting into a helper method that checks that each piece of the chain is present. Check Transformers.java for useful utility methods.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationTransformer.java
```diff
@@ -49,6 +49,7 @@ public Observation apply(CdwObservation cdw) {
      * Specimen reference is omitted since we do not support the a specimen resource and
      * do not want dead links
      */
+    String categoryCode = cdw.getCategory().getCoding().get(0).getCode().value();
```
bschofield-va: `.get(0)` without checking size can blow up

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationTransformer.java
```diff
@@ -64,7 +65,11 @@ public Observation apply(CdwObservation cdw) {
         .valueCodeableConcept(valueCodeableConcept(cdw.getValueCodeableConcept()))
         .interpretation(interpretation(cdw.getInterpretation()))
         .comments(cdw.getComments())
-        .referenceRange(referenceRanges(cdw.getReferenceRanges()))
+        .referenceRange(
```
bschofield-va: Let's move this logic into to `referenceRanges(...)` and update the signature to be `referenceRanges(cdw.getCategory(), cdw.getReferenceRanges())`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationTransformer.java
```diff
@@ -296,7 +291,18 @@ SimpleQuantity referenceRangeQuantity(CdwObservationRefRangeQuantity maybeCdw) {
         .build();
   }
 
-  List<ObservationReferenceRange> referenceRanges(CdwReferenceRanges maybeCdw) {
+  List<ObservationReferenceRange> referenceRanges(
+      CdwReferenceRanges maybeCdw, CdwCategory cdwCategory) {
+    List<CdwCategory.CdwCoding> categoryCoding = cdwCategory.getCoding();
+    if (categoryCoding.size() > 0
+        && !StringUtils.isEmpty(categoryCoding.get(0).getCode().value())
```
eclendenning-va: If getCode() is null, .value() will throw an NPE.

(one more check!)
(one more check!)

PR #209, opened 2019-06-19
Task/adq 69 address fortify concerns

ilaflamme-va:
Addressing the following fortify issues:

Poor Style: Confusing Naming
  The class contains a field and a method with the same name.

Poor Error Handling: Overly Broad Catch
  The catch block handles a broad swath of exceptions, potentially trapping dissimilar issues or problems
  that should not be dealt with at this point in the program.

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/config/MagicReferenceConfig.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/config/MagicReferenceConfig.java
@@ -113,13 +113,11 @@ public class MagicReferenceConfig {
     @Override
     @SneakyThrows
     public void serializeAsField(
-        Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer)
-        throws IOException {
+        Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer) {
       boolean include = true;
       if (writer.getType().getRawClass() == Reference.class
           && writer instanceof BeanPropertyWriter) {
-        Reference value = null;
-        value = (Reference) ((BeanPropertyWriter) writer).get(pojo);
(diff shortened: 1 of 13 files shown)


mr-anderson/src/main/java/gov/va/api/health/mranderson/cdw/impl/SqlResourceRepository.java
```diff
@@ -42,10 +42,19 @@ public SqlResourceRepository(
...
+     * SQL injection.
+     */
+    return "{call [" + schema + "].[" + storedProcedure + "](?,?,?,?,?,?,?)}";
+  }
+
   @Override
   public String execute(Query query) {
     try (Connection connection = Checks.notNull(jdbc.getDataSource()).getConnection()) {
-      try (CallableStatement cs = connection.prepareCall(storedProcedure())) {
+      try (CallableStatement cs = connection.prepareCall(buildStoredProcedure())) {
```
jblefkove-va: The name should indicate that this is a string of SQL. How about `storedProcedureSql()` or `storedProcedureCallSql()` ?

data-query/src/main/java/gov/va/api/health/dataquery/service/config/MagicReferenceConfig.java
```diff
@@ -110,13 +111,15 @@ private boolean hasExtensionField(Object object, String name) {
      * that type of writer so we can peek at the value we are about to serialize.
      */
     @Override
+    @SneakyThrows
     public void serializeAsField(
         Object pojo, JsonGenerator jgen, SerializerProvider provider, PropertyWriter writer)
-        throws Exception {
+        throws IOException {
```
jblefkove-va: Since you added `@SneakyThrows` you should be able to remove `throws XXX` from the signature altogether.

PR #211, opened 2019-06-24
use icn header to do patient reference substitution on procedure reads

eclendenning-va:
Related Story: https://vasdvp.atlassian.net/browse/ADQ-66

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/ProcedureController.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/ProcedureController.java
@@ -4,7 +4,6 @@ import static gov.va.api.health.dataquery.service.controller.Transformers.firstP
 import static gov.va.api.health.dataquery.service.controller.Transformers.hasPayload;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import gov.va.api.health.argonaut.api.resources.Procedure;
 import gov.va.api.health.autoconfig.configuration.JacksonConfig;
@@ -18,14 +17,11 @@ import gov.va.api.health.dataquery.service.mranderson.client.MrAndersonClient;
 import gov.va.api.health.dataquery.service.mranderson.client.Query;
 import gov.va.api.health.dstu2.api.resources.OperationOutcome;
 import gov.va.dvp.cdw.xsd.model.CdwProcedure101Root;
-
-import java.io.Serializable;
 import java.util.Collections;
 import java.util.function.Function;
 import javax.validation.Valid;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.Size;
-
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.SneakyThrows;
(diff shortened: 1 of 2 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/ProcedureController.java
```diff
@@ -102,7 +103,16 @@
 
   /** Read by id. */
   @GetMapping(value = {"/{publicId}"})
-  public Procedure read(@PathVariable("publicId") String publicId) {
+  public Procedure read(
+      @PathVariable("publicId") String publicId,
+      @RequestHeader(value = "X-VA-ICN", required = false) String icnHeader) {
+    if (isNotBlank(icnHeader) && thisLooksLikeAJobForSuperman(icnHeader)) {
```
jblefkove-va: The `isNotBlank` is unnecessary, `thisLooksLikeAJobForSuperman` covers it.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/ProcedureController.java
```diff
@@ -102,7 +103,16 @@
...
+      return usePhoneBooth(
+          transformer.apply(
+              firstPayloadItem(
+                  hasPayload(search(Parameters.forIdentity(publicId)).getProcedures())
+                      .getProcedure())));
```
jblefkove-va: The way this is implemented for the procedure bundle, if you invoke the method with the superman ID, the method invokes itself recursively with the clarkKentId, then passes that through the phone booth.

I haven't tested this, but I think this would work the same way, avoiding the duplication:
```
if (thisLooksLikeAJobForSuperman(icnHeader)) {
  return usePhoneBooth(read(publicId, clarkKentId));
}
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/ProcedureController.java
```diff
@@ -205,6 +215,29 @@ private boolean thisLooksLikeAJobForSuperman(String patient) {
...
   }
 
+  /**
+   * Change a clark-kent procedure into a superman procedure. {@link #clarkKentId} is replaced with
+   * {@link #supermanId} and {@link #clarkKentDisplay} is replaced with {@link #supermanDisplay}
+   *
+   * @see #thisLooksLikeAJobForSuperman(String)
+   */
+  @SneakyThrows
+  private Procedure usePhoneBooth(Procedure clarkKent) {
```
jblefkove-va: This method is effectively identical to the other method that takes a procedure bundle. (All we do is serialize and deserialize the object, and Jackson doesn't care what class it is.) So with a little refactoring you should be able to have one method that take and returns a generic type: `private <T> T usePhoneBooth(T clarkKentObj)`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/ProcedureController.java
eclendenning-va: thisLooksLikeAJobForSuperman (String patient) uses patient.equals which causes a NPE if argument is not checked for blank.

jblefkove-va: image

PR #213, opened 2019-06-25
system.time is now java.time.clock

ilaflamme-va:
Fortify...

### DIFF
--- a/mr-anderson/src/main/java/gov/va/api/health/mranderson/controller/ErrorResponse.java
+++ b/mr-anderson/src/main/java/gov/va/api/health/mranderson/controller/ErrorResponse.java
@@ -1,7 +1,7 @@
 package gov.va.api.health.mranderson.controller;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import java.time.Clock;
+import java.time.Instant;
 import javax.xml.bind.annotation.XmlAccessType;
 import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlRootElement;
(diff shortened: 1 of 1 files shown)


jblefkove-va:
image

mr-anderson/src/main/java/gov/va/api/health/mranderson/controller/ErrorResponse.java
```diff
@@ -26,7 +27,7 @@
   /** Create a new error response based on the given exception. */
   public static ErrorResponse of(@NonNull Exception e) {
     return ErrorResponse.builder()
-        .timestamp(System.currentTimeMillis())
+        .timestamp(Clock.systemDefaultZone().millis())
```
jblefkove-va: I'd prefer `Instant.now().toEpochMilli()`, if Fortify doesn't complain about that, since it is closer to the time classes we regularly use. (I would also add a comment here so that someone doesn't accidentally change it back to `System.currentTimeMillis()` later)

jnguy-va: It seems like Fortify is complaining about `currentTimeMillis()` in community-care:

> The method `currentTimeMillis() is a JNI call.  Improper use of JNI can render Java applications vulnerable to security flaws in other languages.

jhulbert-va:
Jenkins thinks branch is still active and cant merge

jblefkove-va:
Do you want to update the comment before merging?

jhulbert-va:
Reopen to check jenkins no longer thinks branch is active

PR #216, opened 2019-07-03
Feature/adq 100 limit count to 20

bschofield-va:
This PR adds `_count` clamping so that any request with a count higher than a configurable maximum is automatically reduced. By default, the max `_count` is 20. Additionally, integration tests now verify that bundles do not have more than 20 results.

Clamping is achieved by replacing `@RequestParam("_count")` with a new `@CountParameter` annotation.

This also fixes paging links that were potentially wrong when the count was clamped with in the DB.

FYI @jblefkove-va @dtournour-va @alecbert-va @jgarnhart-va @esubocz-va


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/CountParameterResolverTest.java
```diff
@@ -0,0 +1,85 @@
...
+  @SneakyThrows
+  private CountParameter countParameter() {
+    return (CountParameter)
+        getClass().getMethod("reflectMe", Void.class).getParameterAnnotations()[0][0];
+  }
+
+  @Test
+  public void defaultsWhenCountNotSpecified() {
+    useCount(null);
+    assertThat(r.resolveArgument(methodParameter, null, nativeWebRequest, null)).isEqualTo(10);
```
ilaflamme-va: 10, not 15?

data-query-tests/src/test/resources/cdw/index.yaml
```diff
@@ -1,235 +1,235 @@
...
-  count: 2
-- file: AllergyIntoleranceV1.05/AllInt105p185601V825290pg10000ct20.xml
-  query: /AllergyIntolerance:1.05?patient=185601V825290
-  page: 10000
-  count: 20
-- file: AllergyIntoleranceV1.05/AllInt105p185601V825290pg2ct2.xml
-  query: /AllergyIntolerance:1.05?patient=185601V825290
-  page: 2
-  count: 2
+  - file: NoResults.xml
```
jblefkove-va: What triggered the reformatting?

bschofield-va: I had IntelliJ reformat it.

PR #221, opened 2019-07-15
Task/api 1315 replace lablogintest with oauthlogintest

jtorano-va:
Related Story https://vasdvp.atlassian.net/browse/API-1315


data-query-tests/spotbugs-excludes.xml
```diff
@@ -25,17 +25,6 @@
       <Bug pattern="RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE"/>
     </Or>
   </Match>
```
bschofield-va: Thanks for removing!

PR #230, opened 2019-07-29
Datamart Medication Statement model

bschofield-va:
Java to match https://github.freedomstream.io/mulesoft/cdw-schemas/blob/develop/datamart/medicationStatement-json-example.json

See also https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/232

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationstatement/DatamartMedicationStatement.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationstatement/DatamartMedicationStatement.java
@@ -19,8 +19,8 @@ import lombok.NoArgsConstructor;
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class DatamartMedicationStatement {
-  private String objectType;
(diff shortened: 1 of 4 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationstatement/DatamartMedicationStatement.java
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Represents the document returned from Datamart. All fields are required unless specifically
+ * declared as `Optional`. In those cases, the fields will still be populated using
+ * `Optional.empty()`
+ */
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class DatamartMedicationStatement {
```
bschofield-va: @vhaisdharrij08 can you verify the cardinality of these fields. All fields are required unless specifically declared as `Optional<>`

jamesharris1: Medication Statement cardinalities look good.

PR #231, opened 2019-07-29
Datamart: Condition model

bschofield-va:
Model for https://github.freedomstream.io/mulesoft/cdw-schemas/blob/develop/datamart/condition_json_sample.json

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartCondition.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartCondition.java
@@ -1,0 +1,116 @@
+package gov.va.api.health.dataquery.service.controller.condition;
+
+import gov.va.api.health.dataquery.service.controller.datamart.DatamartReference;
+import java.time.Instant;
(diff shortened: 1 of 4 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartCondition.java
```diff
@@ -0,0 +1,116 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class DatamartCondition {
```
bschofield-va: @vhaisdharrij08 Can you review the cardinality of these fields. All fields are required unless specifically declared as `Optional<>`

jamesharris1: Condition cardinality is correct.

PR #232, opened 2019-07-30
Datamart: Medication Statement transformer

bschofield-va:
# Do first https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/230

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationstatement/DatamartMedicationStatementTransformer.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationstatement/DatamartMedicationStatementTransformer.java
@@ -1,0 +1,72 @@
+package gov.va.api.health.dataquery.service.controller.medicationstatement;
+
+import static gov.va.api.health.dataquery.service.controller.Transformers.asDateTimeString;
+import static gov.va.api.health.dataquery.service.controller.Transformers.asReference;
+
+import gov.va.api.health.argonaut.api.resources.MedicationStatement;
+import gov.va.api.health.dstu2.api.datatypes.CodeableConcept;
+import gov.va.api.health.dstu2.api.datatypes.Timing;
+import java.util.List;
+import java.util.Optional;
+import lombok.Builder;
+import lombok.NonNull;
+
+@Builder
+public class DatamartMedicationStatementTransformer {
+
+  @NonNull final DatamartMedicationStatement datamart;
+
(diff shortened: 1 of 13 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationstatement/DatamartMedicationStatementTransformer.java
```diff
@@ -0,0 +1,72 @@
...
+
+  MedicationStatement toFhir() {
+    return MedicationStatement.builder()
+        .resourceType("MedicationStatement")
+        .id(datamart.cdwId())
+        .dateAsserted(asDateTimeString(datamart.dateAsserted()))
+        .dosage(dosage(datamart.dosage()))
+        .effectiveDateTime(asDateTimeString(datamart.effectiveDateTime()))
+        .medicationReference(asReference(datamart.medication()))
+        .note(datamart.note().orElse(null))
```
ilaflamme-va: I like being able to do _.orElse(null)_

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationstatement/DatamartMedicationStatementTransformerTest.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  @Test
+  public void medicationStatement() {
+    assertThat(tx(Datamart.create().medicationStatement()))
+        .isEqualTo(Fhir.create().medicationStatement());
+  }
+
+  @Test
+  public void status() {
```
ilaflamme-va: Should we test the default case: _IllegalArgumentException_ ?

bschofield-va: Impossible to test and actually impossible to hit. However,  should a third status value be created in the future, this IAEx could be fired. I use this technique to help future proof. Consider this alternative implementation to the status(...) transformer method.
```
if (status == null) { return null; }
if (status == DatamartMedicationStatement.active) {
   return MedicationStatement.Status.active
}
return MedicationStatement.Status.complete;
```
Today, this implementation would work equally well as mine, but implementations of this style can let subtle bugs slip in. If `Status.in_progress` were added, and we forgot to update the transformer, my implementation would explode violently. Implementations in the style above would casually return the incorrect value, which could be very hard to detect.

PR #233, opened 2019-07-31
Task/adq 144 co transformer more than meets the eye

bschofield-va:
See also https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/234


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/Transformers.java
```diff
@@ -50,12 +53,30 @@ public static String asDateString(XMLGregorianCalendar maybeDate) {
...
+    return asDateString(maybeDateTime.orElse(null));
+  }
+
+  /** Return null if the date is null, otherwise return an ISO-8601 date. */
+  public static String asDateString(Instant maybeDateTime) {
+    if (maybeDateTime == null) {
+      return null;
+    }
+    return DateTimeFormatter.ofPattern("yyyy-MM-dd", Locale.US)
+        .withZone(ZoneOffset.UTC)
```
jblefkove-va: They're about the same complexity, but I think `LocalDateTime.ofInstant(maybeDateTime, ZoneOffset.UTC)` would be cleaner.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/Transformers.java
```diff
@@ -50,12 +53,30 @@ public static String asDateString(XMLGregorianCalendar maybeDate) {
     return formatter.format(maybeDate.toGregorianCalendar().getTime());
   }
 
+  /** Return null if the date is null, otherwise return an ISO-8601 date. */
+  public static String asDateString(Optional<Instant> maybeDateTime) {
+    if (maybeDateTime == null) {
```
jblefkove-va: Doesn't null-checking an Optional defeat its purpose?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartConditionTransformer.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+    throw new IllegalArgumentException("Unsupported ICD code version: " + icdCode.version());
+  }
+
+  /** Convert the datamart structure to FHIR compliant structure. */
+  public Condition toFhir() {
+    return Condition.builder()
+        .resourceType("Condition")
+        .abatementDateTime(asDateTimeString(datamart.abatementDateTime()))
+        .asserter(asReference(datamart.asserter().orElse(null)))
```
jblefkove-va: Should there be another `Transformers.asReference` method that takes `Optional<DatamartReference>` ?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartConditionTransformer.java
```diff
@@ -0,0 +1,135 @@
...
+                Coding.builder()
+                    .system(systemOf(icdCode))
+                    .code(icdCode.code())
+                    .display(icdCode.display())
+                    .build()))
+        .build();
+  }
+
+  private String systemOf(IcdCode icdCode) {
+    if ("10".equals(icdCode.version())) {
```
jblefkove-va: Should there be a null check on `icdCode` as well?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartConditionTransformer.java
```diff
@@ -0,0 +1,135 @@
...
+      return "http://hl7.org/fhir/sid/icd-10";
+    }
+    if ("9".equals(icdCode.version())) {
```
jblefkove-va: This could be a little more compact as a switch.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartConditionTransformer.java
bschofield-va: In the DatamartCondition, the `icdModel` field can be null, but that check is done before this method is called.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/Transformers.java
bschofield-va: It's probably overkill, but I was protecting against an Optional field that is missing a lazy getter

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/Transformers.java
bschofield-va: `LocalDateTime` ends up printing `2007-12-03T10:15:30`, I need to limit it just the date.

jblefkove-va: Sorry, I meant `LocalDate`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartConditionTransformer.java
bschofield-va: If `icdCode.version()` is null, this approach will throw a IAX with a nice message. The `switch` would throw an NPE, unless I added an explicit null check (which would make it more verbose)

PR #234, opened 2019-07-31
Task/adq 142 co jpa

bschofield-va:
### Do this first: https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/233

Datamart Condition entity model and read /raw support
Also introduces `Parameters.identifierOf(...)` utility method

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionEntity.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionEntity.java
@@ -1,0 +1,69 @@
+package gov.va.api.health.dataquery.service.controller.condition;
+
+import gov.va.api.health.autoconfig.configuration.JacksonConfig;
+import javax.persistence.Basic;
+import javax.persistence.Column;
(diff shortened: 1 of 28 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionEntity.java
```diff
@@ -0,0 +1,69 @@
...
+ *  [AsserterCDWId] [int] NULL,
+ *  [Category] [varchar](50) NULL,
+ *  [ClinicalStatus] [varchar](50) NULL,
+ *  [DateRecorded] [datetime2](0) NULL,
+ *  [EncounterCDWId] [bigint] NULL,
+ *  [OnSet] [datetime2](0) NULL,
+ *  [Condition] [varchar](max) NULL,
+ *  [ETLBatchId] [int] NULL,
+ *  [ETLCreateDate] [datetime2](0) NULL,
+ *  [ETLEditDate] [datetime2](0) NULL,
```
jblefkove-va: Your fake SQL is missing a closing paren. 😃

PR #237, opened 2019-08-02
Remove Date field from Response

jjennings-va:
https://vasdvp.atlassian.net/browse/API-1411

Stop mapping optional field of Recorded Date to response


bschofield-va:
# Please do not merge yet

PR #238, opened 2019-08-03
Task/adq 145 co controller

bschofield-va:
This is ready now.

Adds Datamart support for Condition resources.
New application property `datamart.condition` controls default behavior. `false` uses CDW, `true` uses Datamart. By default, `false`.

Unlike Patient and DR, this resource leverages Spring Data repositories for JPA/Entity Manager.
Note that paging with Spring data starts at `0` but our paging starts at 1.


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/WitnessProtection.java
```diff
@@ -24,6 +35,39 @@
 public class WitnessProtection {
   private IdentityService identityService;
 
+  private <T extends HasReplaceableId> Function<T, Stream<DatamartReference>> imbellish(
```
jblefkove-va: embellish

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionController.java
```diff
@@ -185,4 +243,143 @@ public OperationOutcome validate(@RequestBody Condition.Bundle bundle) {
...
+    }
+
+    private Bundle bundle(
+        MultiValueMap<String, String> parameters, int count, Page<ConditionEntity> entities) {
+
+      log.info("Search {} found {} results", parameters, entities.getTotalElements());
+      if (count == 0) {
+        return bundle(parameters, emptyList(), (int) entities.getTotalElements());
+      }
+      List<ConditionEntity> xxx = entities.get().collect(Collectors.toList());
```
jblefkove-va: xxx!

bschofield-va: dang it! That was some debugging cruft

PR #242, opened 2019-08-05
JPA entity model and readRaw

ilaflamme-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationstatement/MedicationStatementController.java
```diff
@@ -78,6 +83,17 @@ public MedicationStatement read(@PathVariable("publicId") String publicId) {
                 .getMedicationStatement()));
   }
 
+  /** Read by id, raw data. */
+  @GetMapping(value = {"/{publicId}/raw"})
+  public String readRaw(@PathVariable("publicId") String publicId) {
+    MultiValueMap<String, String> publicParameters = Parameters.forIdentity(publicId);
```
bschofield-va: Do this like `Condition.Datamart.readRaw(...)`

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationstatement/DatamartMedicationStatementControllerTest.java
```diff
@@ -0,0 +1,75 @@
...
+        repository);
+  }
+
+  public void mockMedicationStatementIdentity(String publicId, String cdwId) {
+    when(ids.lookup(publicId))
+        .thenReturn(
+            List.of(
+                ResourceIdentity.builder()
+                    .system("CDW")
+                    .resource("CONDITION")
```
jblefkove-va: copy-paste error

PR #248, opened 2019-08-06
Task/adq 151 im controller

bschofield-va:
Datamart immunization controller, transformer, and integration tests


jblefkove-va:
image

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -106,7 +106,7 @@ private static TestIds localIds() {
         .diagnosticReport("1400096309217:L")
         .encounter("1200753214085")
         .diagnosticReports(diagnosticReports())
-        .immunization("1000000043979")
+        .immunization("12345678")
```
jblefkove-va: I've got the same combination on my luggage!

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunizationTransformer.java
```diff
@@ -0,0 +1,94 @@
...
+  private final DatamartImmunization datamart;
+
+  List<Annotation> note(Optional<String> note) {
+    return note.isPresent() ? List.of(Annotation.builder().text(note.get()).build()) : null;
+  }
+
+  List<Immunization.Reaction> reaction(Optional<DatamartReference> reaction) {
+    return reaction.isEmpty()
+        ? null
+        : List.of(Immunization.Reaction.builder().detail(asReference(reaction)).build());
```
jblefkove-va: Is it safe to assume that if `reaction` is not empty, all the reference fields are populated and `asReference()` is guaranteed to return something?

data-query/src/main/resources/application.properties
```diff
@@ -56,5 +56,6 @@ datamart.allergy-intolerance=false
 datamart.condition=false
 datamart.immunization=false
 datamart.diagnostic-report=true
+datamart.immunization=false
```
jblefkove-va: Repeated

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunizationTransformer.java
bschofield-va: The _assumption_ is that if reaction is present then at least _display_ or, _type_ and _reference_ fields will be set.

bschofield-va: I added some logic to help protect from an empty, but assigned reaction.

PR #253, opened 2019-08-07
Task/adq 164 pr doc model

bschofield-va:
Do this first: https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/252

Adds Data query procedure model and transformer


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/DatamartProcedureTransformer.java
```diff
@@ -0,0 +1,45 @@
...
+
+  List<CodeableConcept> reasonNotPerformed(Optional<String> reasonNotPerformed) {
+    if (reasonNotPerformed.isEmpty()) {
+      return null;
+    }
+    return List.of(CodeableConcept.builder().text(reasonNotPerformed.get()).build());
+  }
+
+  Status status(DatamartProcedure.Status status) {
+    return EnumSearcher.of(Procedure.Status.class).find(status.toString());
```
jblefkove-va: Is `status` guaranteed non-null?

bschofield-va: yes

PR #254, opened 2019-08-07
Task/api 1440 condition code

jhulbert-va:
[API-1440](https://vasdvp.atlassian.net/browse/API-1440)

* Does not print code for conditions if `.coding.display` or `.coding.code` is null 
* Adds `.text` field that matching the `.coding.display`

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartCondition.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartCondition.java
@@ -110,10 +110,7 @@ public class DatamartCondition implements HasReplaceableId {
 
     /** Determine if an icd code can be mapped without null values. */
     public boolean isUsable() {
-      if (code != null && display != null) {
-        return true;
-      }
-      return false;
+      return code != null && display != null;
     }
   }
 
@@ -128,10 +125,7 @@ public class DatamartCondition implements HasReplaceableId {
 
     /** Determine if a snomed code can be mapped without null values. */
     public boolean isUsable() {
-      if (code != null && display != null) {
-        return true;
-      }
-      return false;
+      return code != null && display != null;
     }
   }
 }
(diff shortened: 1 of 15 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartConditionTransformer.java
```diff
@@ -27,10 +27,14 @@
    */
   CodeableConcept bestCode() {
     if (datamart.snomed().isPresent()) {
-      return code(datamart.snomed().get());
+      if (datamart.snomed().get().code() != null && datamart.snomed().get().display() != null) {
```
bschofield-va: Can we make it pretty?
```
if (datamart.snomed().isPresent() && snomed.get().isUsuable()) {
  return code( ... );
}
...
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartConditionTransformer.java
```diff
@@ -27,10 +27,14 @@
    */
   CodeableConcept bestCode() {
```
bschofield-va: Update the transformer unit tests to verify new code paths

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/DatamartCondition.java
```diff
@@ -101,18 +102,36 @@
...
+  @JsonIgnoreProperties({"usable"})
   public static class IcdCode {
     String code;
     String display;
     String version;
+
+    /** Determine if an icd code can be mapped without null values. */
+    public boolean isUsable() {
+      if (code != null && display != null) {
+        return true;
```
jblefkove-va: This could be simplified to `return code != null && display != null`

PR #255, opened 2019-08-07
Task/adq 179 medication datamart transformer

eclendenning-va:
Related Story : https://vasdvp.atlassian.net/browse/ADQ-179


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/DatamartMedicationTransformer.java
```diff
@@ -0,0 +1,53 @@
...
+        .build();
+  }
+
+  /** Convert the datamart structure to FHIR compliant structure. */
+  public Medication toFhir() {
+    return Medication.builder()
+        .resourceType(Medication.class.getSimpleName())
+        .id(datamart.cdwId())
+        .product(product(datamart.product()))
+        .text(text(datamart.rxnorm().text))
```
bschofield-va: `text()` use function accessor ... in fact, that data structure should have _private_ members.

PR #257, opened 2019-08-07
Task/adq 185 immunization entity model

jhulbert-va:
[ADQ-185](https://vasdvp.atlassian.net/browse/ADQ-185)

* Change entity model for Immunization for `CDWId` column to string

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/ImmunizationController.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/ImmunizationController.java
@@ -54,8 +54,9 @@ import org.springframework.web.bind.annotation.RestController;
 @Validated
 @RestController
 @RequestMapping(
-    value = {"Immunization", "/api/Immunization"},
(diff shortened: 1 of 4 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/ImmunizationController.java
```diff
@@ -244,9 +243,12 @@ private Bundle bundle(
     }
 
     ImmunizationEntity findById(String publicId) {
-      Optional<ImmunizationEntity> entity =
-          repository.findById(new BigInteger(witnessProtection.toCdwId(publicId)));
-      return entity.orElseThrow(() -> new NotFound(publicId));
+      String cdwId = witnessProtection.toCdwId(publicId);
+      Optional<ImmunizationEntity> maybeEntity = repository.findById(cdwId);
+      if (!maybeEntity.isPresent()) {
```
bschofield-va: Why did we turn 1 line into 4?

PR #259, opened 2019-08-07
Task/adq 165 pr jpa model

bschofield-va:
Introduces Criteria API for complex searches. Classes of interest:
- ProcedureRepository
- DateTimeParameters
- ProcedureController.Datamart

### WARNING

`Procedure` is a key word for SqlServer. I've had to do some refactoring to make it work.
- Schema and table name are now explicitly defined on entities
- By default, Hibernate is now configured to quote table and column names ... but this doesn't work for H2 database used for unit and integration tests, so it has to be explicitly disabled for tests.

```
sed -i 's/name = "app.\([A-Za-z]\+\)"/name = "\1", schema = "app"/' $(git grep -l '@Table')
```


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/ProcedureController.java
```diff
@@ -224,4 +289,140 @@ public OperationOutcome validate(@RequestBody Procedure.Bundle bundle) {
...
+    }
+
+    String readRaw(String publicId) {
+      return findById(publicId).payload();
+    }
+
+    Collection<DatamartProcedure> replaceReferences(Collection<DatamartProcedure> resources) {
+      /*
+       * Reaction is not a reference to another resource and is intentionally excluded from
+       * registration.
```
jblefkove-va: Is this comment leftover from another resource?

bschofield-va: yup

PR #267, opened 2019-08-12
Support for _count=0 for Datamart resources

bschofield-va:
This PR addresses (at least some) of the QA integration test failures


bschofield-va:
@jhulbert-va @jblefkove-va You'll need to update your controllers

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/AllergyIntoleranceController.java
```diff
@@ -278,7 +278,7 @@ void replaceReferences(Collection<DatamartAllergyIntolerance> resources) {
       int page = Parameters.pageOf(publicParameters);
       int count = Parameters.countOf(publicParameters);
       Page<AllergyIntoleranceEntity> entitiesPage =
-          repository.findByIcn(cdwIcn, PageRequest.of(page - 1, count));
+          repository.findByIcn(cdwIcn, PageRequest.of(page - 1, count == 0 ? 1 : count));
```
jblefkove-va: Won't this return a bundle with one element, not an empty bundle?

PR #272, opened 2019-08-13
Bug/adq 186 allergy intolerance manifestation coding

jhulbert-va:
[ADQ-186](https://vasdvp.atlassian.net/browse/ADQ-186)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/DatamartAllergyIntoleranceTransformer.java
```diff
@@ -39,18 +39,15 @@
     return EnumSearcher.of(AllergyIntolerance.Certainty.class).find(certainty.toString());
   }
 
-  private List<CodeableConcept> manifestations(List<DatamartCoding> manifestations) {
-    if (isEmpty(manifestations)) {
+  private List<CodeableConcept> manifestations(List<DatamartCoding> maybeManifestations) {
```
jblefkove-va: Why the name change? The `maybe` prefix is usually for `Optional<>`s

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/DatamartAllergyIntoleranceTransformer.java
```diff
@@ -39,18 +39,15 @@
...
-        CodeableConcept.builder()
-            .coding(
-                manifestations
-                    .stream()
-                    .map(m -> asCoding(Optional.ofNullable(m)))
-                    .collect(Collectors.toList()))
-            .build());
+    List<CodeableConcept> manifestations = new ArrayList<>();
+    for (DatamartCoding manifestation : maybeManifestations) {
+      manifestations.add(CodeableConcept.builder().coding(asList(asCoding(manifestation))).build());
```
jblefkove-va: The old code wasn't doing this either, but looking at it now, there should be a check here to avoid creating `CodeableConcept`s when `asCoding()` returns `null`.

You can do the conversion and the null-check all in one stream. Map each `DatamartCoding` to `asCoding()`, then `filter(Objects::nonNull)`, then map again to build each `CodeableConcept`.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/DatamartAllergyIntoleranceTransformer.java
```diff
@@ -39,15 +40,20 @@
...
+    List<Coding> codings =
+        manifestations
+            .stream()
+            .map(m -> asCoding(m))
+            .filter(Objects::nonNull)
+            .collect(Collectors.toList());
+    return codings
+        .stream()
+        .map(coding -> CodeableConcept.builder().coding(asList(coding)).build())
+        .collect(Collectors.toList());
```
jblefkove-va: Much improved 👍 

Don't forget `emptyToNull()` at the end

PR #274, opened 2019-08-14
Datamart components now have better logging (using @Loggable)

bschofield-va:
# ATTENTION
Do not merge until parent pom versions have been updated.
See https://github.com/department-of-veterans-affairs/health-apis-parent/pull/28


data-query/src/main/java/gov/va/api/health/dataquery/service/healthcheck/SteelThreadSystemCheck.java
```diff
@@ -91,7 +91,6 @@ public void runSteelThreadCheckAsynchronously() {
     if ("skip".equals(id)) {
       return;
     }
-    log.info("Performing health check.");
```
jblefkove-va: 🚮

PR #281, opened 2019-08-15
Added -Ddatamart=true support to the crawler 

bschofield-va:
which will add the -HDatamart:<value> header if specified to all requests.


bschofield-va:
yolo

PR #300, opened 2019-09-09
Task/adq 237 ids properties file

jhulbert-va:
[ADQ-237](https://vasdvp.atlassian.net/browse/ADQ-237)

* RevealSecretIdentity.java now uses properties file, not local ids
* Transformers use RSI object to unmask ids
* Shell script passes in properties file location of `health-apis-data-query-synthetic-records/identity-service.properties` (can be overwritten)


jhulbert-va:
Merged Branch into `task/ADQ-235-dqlserver-db` branch

PR #334, opened 2019-10-01
Refactor Observation Unit Tests

ilaflamme-va:
In preparation for the last_updated changes approaching, refactor the observation unit tests and sample data.


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/observation/DatamartObservationControllerTest.java
```diff
@@ -1,421 +1,524 @@
...
+                    link(LinkRelation.self, "http://fonzy.com/cool/Observation?patient=p0", 1, 10),
+                    link(
+                        LinkRelation.last,
+                        "http://fonzy.com/cool/Observation?patient=p0",
+                        1,
+                        10))));
   }
 
   @Test
   public void searchByPatientAndCategory() {
```
bschofield-va: I think we need a _search by categories_ `categories=laboratory,vital-signs`

ilaflamme-va: Addressed with a new test.

PR #346, opened 2019-10-15
Task/adq 336 condition patient matching

jhulbert-va:
[ADQ-336](https://vasdvp.atlassian.net/browse/ADQ-336)

* `X-VA-INCLUDE-ICN` header available in Condition Response


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionIncludesIcnMajig.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+@ControllerAdvice
+public class ConditionIncludesIcnMajig
+    extends AbstractIncludesIcnMajig<Condition, Condition.Entry, Condition.Bundle> {
+  /** Converts the reference to a Datamart Reference to pull out the patient id. */
+  public ConditionIncludesIcnMajig() {
+    super(
+        Condition.class,
+        Condition.Bundle.class,
+        body -> Stream.of(Transformers.asReferenceId(body.patient())).filter(Objects::nonNull));
```
bschofield-va: put this filter in the abstract parent so its only done once

PR #385, opened 2019-11-08
bug/SHANK-137-openapi-unavailable

jhulbert-va:
https://vasdvp.atlassian.net/browse/SHANK-137

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/DataQueryHomeController.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/DataQueryHomeController.java
@@ -39,7 +39,7 @@ public class DataQueryHomeController {
    * redirect.
    */
   @GetMapping(
-    value = {"/", "/openapi.json", "/api/openapi.json"},
+    value = {"dstu2/", "dstu2/openapi.json"},
     produces = "application/json"
   )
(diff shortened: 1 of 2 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/DataQueryHomeController.java
```diff
@@ -39,7 +39,7 @@ public String openapiContent() throws IOException {
    * redirect.
    */
   @GetMapping(
-    value = {"/", "/openapi.json", "/api/openapi.json"},
+    value = {"dstu2/", "dstu2/openapi.json"},
```
bschofield-va: This means that `/` will now return a `404`. No one really uses `/` so it's not a big deal ...

jhulbert-va: Shouldn't it be `dstu2/` now? not just `/`?

bschofield-va: yes it should ... but now what _should_ the behavior of this application be when someone hits the root? Which can be decided later... can you drop an item in the back log so we don't forget?

PR #393, opened 2019-11-18
SSS-65 MitreMinimartMaker location support

jblefkove-va:
https://vasdvp.atlassian.net/browse/SSS-65

Add support to `MitreMinimartMaker` for location resource

see also synthetic data PR: https://github.com/department-of-veterans-affairs/health-apis-data-query-synthetic-records/pull/10

payload format: https://github.freedomstream.io/mulesoft/cdw-schemas/pull/69


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/DatamartLocation.java
```diff
@@ -0,0 +1,81 @@
...
+    active,
+    inactive
+  }
+
+  @Data
+  @Builder
+  @NoArgsConstructor(access = AccessLevel.PRIVATE)
+  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  public static class Address {
+    private String line1;
```
jhulbert-va: Will there ever be more than one line? Most of the addresses we've seen in the past allow at least 3 lines.

jblefkove-va: According to James Harris, it's just one line for the street address.

PR #397, opened 2019-11-27
SHANK-153 - Generic OO Example

beaugrantham-va:
https://vasdvp.atlassian.net/browse/SHANK-153

Note that this PR uses SNAPSHOT versions.

These will be updated when accompanying PR is merged.

See also https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/pull/37


bschofield-va:
This PR is not needed ... Jenkins will automatically upgrade the dependency the next time it builds.

PR #401, opened 2019-12-03
Feature/shank 6 local mitre db

bschofield-va:
Local Mitre DB has been rebuilt with 4 patients from the DQ Synthetic Database.


data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/ResourceVerifier.java
```diff
@@ -56,19 +46,12 @@
   private final Set<Class<?>> verifiedPageBoundsClasses =
       Collections.newSetFromMap(new ConcurrentHashMap<>());
 
-  private ImmutableList<Class<?>> DATAMART_RESOURCES =
+  private ImmutableList<Class<?>> DATAMART_AND_CDW_RESOURCES =
```
bschofield-va: @jblefkove-va As Location, Organization, etc. are brought over they will need to be included in this list so that both CDW stored procedure and datamart versions are tested

data-query/src/test/java/gov/va/api/health/dataquery/tools/DatamartExporter.java
```diff
@@ -29,29 +35,28 @@
...
-          ProcedureEntity.class
-          //
-          );
+          ExportForPatientCriteria.of(AllergyIntoleranceEntity.class),
+          ExportForPatientCriteria.of(ConditionEntity.class),
+          ExportForPatientCriteria.of(DiagnosticReportCrossEntity.class),
+          ExportForPatientCriteria.of(DiagnosticReportsEntity.class),
+          ExportForPatientCriteria.of(ImmunizationEntity.class),
+          ExportForPatientCriteria.of(MedicationOrderEntity.class),
+          ExportAllCriteria.of(MedicationEntity.class),
```
bschofield-va: @jblefkove-va Resources that are not specific to a patient, e.g. Location, can be exported like this to get all records.

PR #408, opened 2019-12-06
SSS-69 refactor practitioner for datamart

jjennings-va:
https://vasdvp.atlassian.net/browse/SSS-69

Adding a few more tests to up code coverage, rest of PR is ready for review


data-query-tests/rebuild-datamart-index.sh
```diff
@@ -34,6 +34,13 @@ do
     -e 's/identifier=1400000140034/identifier=1000001259996/' \
     -e 's/\(MedicationOrder.*\)1200389904206/\11400162277477/' \
     -e 's/1201051417263:V/1201472002950:V/' \
+    -e 's/identifier=10092125/identifier=416704/' \
+    -e 's/Appointment:1.01/Appointment:1.02/' \
+    -e 's/Encounter:1.01/Encounter:1.02/' \
+    -e 's/Location:1.00/Location:1.02/' \
+    -e 's/MedicationDispense:1.00/MedicationDispense:1.02/' \
+    -e 's/Organization:1.00/Organization:1.02/' \
+    -e 's/Practitioner:1.00/Practitioner:1.02/' \
```
jblefkove-va: Since you edited `datamart-index.yaml` by hand, let's revert these changes

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerController.java
```diff
@@ -71,13 +115,29 @@
...
   }
 
+  /** Read by id. */
+  @GetMapping(
+    value = {"/{publicId}"},
+    headers = {"raw=true"}
+  )
+  public String readRaw(@PathVariable("publicId") String publicId, HttpServletResponse response) {
+    PractitionerEntity entity = datamart.readRaw(publicId);
+    AbstractIncludesIcnMajig.addHeader(response, entity.npi());
```
jblefkove-va: I think this should be `AbstractIncludesIcnMajig.addHeaderForNoPatients(response);` since practitioners don't have ICNs

data-query/src/main/resources/application.properties
```diff
@@ -63,5 +64,6 @@ datamart.medication-statement=true
 datamart.observation=true
 datamart.patient=true
 datamart.procedure=true
+datamart.practitioner=true
```
jblefkove-va: False for now

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/DatamartPractitionerTransformer.java
```diff
@@ -0,0 +1,216 @@
...
+      return null;
+    }
+    return results.stream().map(s -> healthCareService(s)).collect(Collectors.toList());
+  }
+
+  private Reference location(DatamartReference source) {
+    if (source == null) {
+      return null;
+    }
+    return asReference(source.type(Optional.of("Location")));
```
jblefkove-va: The `replaceReferences` method in the controller will also fail to process references that are missing types. Move this into `PractitionerEntity.asDatamartPractitioner()` to match how location does it. (Ditto for line 99.)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/DatamartPractitionerTransformer.java
```diff
@@ -0,0 +1,195 @@
+package gov.va.api.health.dataquery.service.controller.practitioner;
```
bschofield-va: Should be `Dstu2PractitionerTransformer`

data-query/src/main/resources/application.properties
```diff
@@ -63,5 +64,6 @@ datamart.medication-statement=true
 datamart.observation=true
 datamart.patient=true
 datamart.procedure=true
+datamart.practitioner=false
```
bschofield-va: swap 66 and 67

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/DatamartPractitionerSamples.java
```diff
@@ -0,0 +1,168 @@
...
+                  DatamartPractitioner.Telecom.builder()
+                      .use(DatamartPractitioner.Telecom.Use.mobile)
+                      .system(DatamartPractitioner.Telecom.System.phone)
+                      .value("123-456-1234")
+                      .build()))
+          .build();
+    }
+
+    @AllArgsConstructor(staticName = "create")
+    static class Fhir {
```
bschofield-va: Rename to `Dstu2`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerController.java
```diff
@@ -128,4 +190,102 @@ public OperationOutcome validate(@RequestBody Practitioner.Bundle bundle) {
...
+      return bundler.bundle(
+          Bundler.BundleContext.of(
+              linkConfig,
+              reports,
+              Function.identity(),
+              Practitioner.Entry::new,
+              Practitioner.Bundle::new));
+    }
+
+    Practitioner.Bundle bundle(
```
bschofield-va: This method is never being called

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerController.java
```diff
@@ -128,4 +190,102 @@ public OperationOutcome validate(@RequestBody Practitioner.Bundle bundle) {
...
+      }
+      return BooleanUtils.isTrue(BooleanUtils.toBooleanObject(datamartHeader));
+    }
+
+    PageRequest page(int page, int count) {
+      return PageRequest.of(page - 1, count == 0 ? 1 : count, PractitionerEntity.naturalOrder());
+    }
+
+    Practitioner read(String publicId) {
+      return transform(findById(publicId).asDatamartPractitioner());
```
bschofield-va: Read is not properly replacing references

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/DatamartPractitionerTransformerTest.java
```diff
@@ -0,0 +1,85 @@
+package gov.va.api.health.dataquery.service.controller.practitioner;
```
bschofield-va: this is missing several empty/ null cases

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerController.java
bschofield-va: still does not work

bschofield-va:
```
$ curl -HDatamart:true http://localhost:8090/dstu2/Practitioner/I2-DQEIIZRKOJELRZI5JGTCDH342JIFVNCMINIHL44LHJSJEKMCZFJQ0000 | jq '.practitionerRole[]|.managingOrganization,.location'
{
  "reference": "http://localhost:8090/dstu2/Organization/561596:I",
  "display": "CHEYENNE VA MEDICAL"
}
[
  {
    "reference": "http://localhost:8090/dstu2/Location/43817:L",
    "display": "CHEY MEDICAL"
  }
]
```

bschofield-va: Searches also do work
```
curl -HDatamart:true http://localhost:8090/dstu2/Practitioner?_id=I2-DQEIIZRKOJELRZI5JGTCDH342JIFVNCMINIHL44LHJSJEKMCZFJQ0000
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2Practitioner.java
```diff
@@ -17,7 +17,7 @@
 @Builder
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
-public class DatamartPractitioner implements HasReplaceableId {
+public class Dstu2Practitioner implements HasReplaceableId {
```
jhulbert-va: Pretty sure this should still be `DatamartPractitioner.java`. It reflects the Practitioner Datamart schema.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerController.java
```diff
@@ -34,19 +48,46 @@
  * Request Mappings for Practitioner Profile, see http://hl7.org/fhir/DSTU2/practitioner.html for
  * implementation details.
  */
-@SuppressWarnings("WeakerAccess")
+@Slf4j
```
jblefkove-va: This logger is unused

data-query/spotbugs-excludes.xml
```diff
@@ -29,8 +29,8 @@
       <Class name="gov.va.api.health.dataquery.service.controller.observation.DatamartObservation"/>
       <Class name="gov.va.api.health.dataquery.service.controller.organization.DatamartOrganization"/>
       <Class name="gov.va.api.health.dataquery.service.controller.patient.DatamartPatient"/>
-      <Class name="gov.va.api.health.dataquery.service.controller.practitioner.DatamartPractitioner"/>
-      <Class name="gov.va.api.health.dataquery.service.controller.practitioner.DatamartPractitioner$PractitionerRole"/>
+      <Class name="gov.va.api.health.dataquery.service.controller.practitioner.Dstu2Practitioner"/>
```
jblefkove-va: As Josh mentioned, this should still be `DatamartPractitioner` since it is the datamart model class

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerController.java
```diff
@@ -34,19 +48,46 @@
...
 @Validated
 @RestController
+@SuppressWarnings("WeakerAccess")
 @RequestMapping(
   value = {"/dstu2/Practitioner"},
   produces = {"application/json", "application/json+fhir", "application/fhir+json"}
 )
-@AllArgsConstructor(onConstructor = @__({@Autowired}))
 public class PractitionerController {
```
jblefkove-va: This should be renamed to `Dstu2PractitionerController`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
```diff
@@ -50,7 +51,21 @@ static Sort naturalOrder() {
   }
 
   @SneakyThrows
-  DatamartPractitioner asDatamartPractitioner() {
-    return JacksonConfig.createMapper().readValue(payload, DatamartPractitioner.class);
+  Dstu2Practitioner asDatamartPractitioner() {
+    Dstu2Practitioner dm = JacksonConfig.createMapper().readValue(payload, Dstu2Practitioner.class);
+
+    if (dm.practitionerRole().get().managingOrganization() != null
```
jblefkove-va: For this check and the other below, you need to verify `dm.practitionerRole().isPresent()` first. Also, `managingOrganization` is an optional, so you don't need to check if it's null, but you should check if its present. So the logic is, "if practitioner role is present and managing organization is present and managing organization type is empty, overwrite it"

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
```diff
@@ -50,7 +51,21 @@ static Sort naturalOrder() {
...
+        && dm.practitionerRole().get().managingOrganization().get().type().isEmpty()) {
+      // Hack... make sure reference type is populated
+      dm.practitionerRole().get().managingOrganization().get().type(Optional.of("Organization"));
+    }
+
+    if (dm.practitionerRole().get().location() != null
+        && dm.practitionerRole().get().location().get(0).type().isEmpty()) {
```
jblefkove-va: This check should be a loop over all the locations, not just the first one

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerControllerTest.java
```diff
@@ -0,0 +1,185 @@
...
+@RunWith(SpringRunner.class)
+public class Dstu2PractitionerControllerTest {
+
+  HttpServletResponse response = mock(HttpServletResponse.class);
+
+  private IdentityService ids = mock(IdentityService.class);
+
+  @Autowired private PractitionerRepository repository;
+
+  @Autowired private TestEntityManager entityManager;
```
jblefkove-va: This field is unused

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
bschofield-va: Furthermore, the presence of a list does not guarantee the list is populated and the above condition can throw a NullPointerException

```
var x = List.of();
if (x != null && x.get(0).isAwesome) {
   ...
}
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerController.java
```diff
@@ -128,4 +187,79 @@ public OperationOutcome validate(@RequestBody Practitioner.Bundle bundle) {
...
+
+    Practitioner read(String publicId) {
+      return transform(findById(publicId).asDatamartPractitioner());
+    }
+
+    PractitionerEntity readRaw(String publicId) {
+      return findById(publicId);
+    }
+
+    private Collection<Dstu2Practitioner> replaceReferences(
```
jblefkove-va: Related to Bryan's comment, your IDE should be warning you that this method is unused

jblefkove-va:
FYI reference replacement can be verified in the unit tests. Take another look at `Dstu2LocationControllerTest` and how the tests use the `addMockIdentities()` method. You should structure the tests with pairs of public/private IDs for each reference object. For a test practitioner with one location, that is six total: public and private/CDW ID for the practitioner, the managing organization, and the location.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerController.java
```diff
@@ -0,0 +1,265 @@
...
+
+    boolean isDatamartRequest(String datamartHeader) {
+      if (StringUtils.isBlank(datamartHeader)) {
+        return defaultToDatamart;
+      }
+      return BooleanUtils.isTrue(BooleanUtils.toBooleanObject(datamartHeader));
+    }
+
+    Practitioner read(String publicId) {
+      DatamartPractitioner location = findById(publicId).asDatamartPractitioner();
```
jblefkove-va: name is off

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+  static String birthDate(Optional<LocalDate> source) {
+    LocalDate date = source.isEmpty() ? null : source.get();
+    if (date == null) {
+      return null;
+    }
+    return date.toString();
+  }
+
+  static Reference healthCareService(String s) {
+    return Reference.builder().display(s).build();
```
jblefkove-va: Should this return null if the string is blank?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  static List<Reference> healthcareServices(Optional<String> results) {
```
jblefkove-va: better parameter name?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+    if (results.isEmpty()) {
+      return null;
+    }
+    return results.stream().map(s -> healthCareService(s)).collect(Collectors.toList());
```
jblefkove-va: should this use `emptyToNull()` in case `healthCareService()` returns null?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  static HumanName name(DatamartPractitioner.Name source) {
+    if (source == null
+        || allBlank(source.family(), source.given(), source.prefix(), source.suffix())) {
+      return null;
+    }
+    return convert(
```
jblefkove-va: I don't think `convert()` is providing any benefit here vs. just doing `return HumanName.builder()...` directly

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+        source,
+        name ->
+            HumanName.builder()
+                .family(nameList(name.family()))
+                .given(nameList(name.given()))
+                .suffix(nameList(name.prefix().isEmpty() ? null : name.suffix().get()))
```
jblefkove-va: Instead of this inline conditional, how about making `nameList` accept the `Optional<String>` instead?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+
+  static List<String> nameList(String source) {
+    if (isBlank(source)) {
+      return null;
+    }
+    return singletonList(source);
+  }
+
+  static CodeableConcept role(Optional<DatamartCoding> source) {
+    if (source == null || source.get().code().get() == null) {
```
jblefkove-va: missing check for empty code

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+    if (source == null
+        || allBlank(
+            source.healthCareService(),
+            source.location(),
+            source.managingOrganization(),
+            source.role())) {
+      return null;
+    }
+    return Practitioner.PractitionerRole.builder()
+        .location(singletonList(asReference(datamart.practitionerRole().get().location().get(0))))
```
jblefkove-va: This appears to populate the list with only the first location instead of all of them

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
```diff
@@ -45,12 +45,27 @@
   @Column(name = "Practitioner")
   private String payload;
 
-  static Sort naturalOrder() {
```
jblefkove-va: Did you mean to remove this?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
```diff
@@ -45,12 +45,27 @@
...
+        JacksonConfig.createMapper().readValue(payload, DatamartPractitioner.class);
+
+    if (dm.practitionerRole().isPresent()
+        && dm.practitionerRole().get().managingOrganization().isPresent()
+        && dm.practitionerRole().get().managingOrganization().get().type().isEmpty()) {
+      // Hack... make sure reference type is populated
+      dm.practitionerRole().get().managingOrganization().get().type(Optional.of("Organization"));
+    }
+
+    if (dm.practitionerRole().isPresent() && dm.practitionerRole().get().location() != null) {
```
jblefkove-va: `dm.practitionerRole().get().location()` is a lazily-initialized collection, so you don't need to check that for null

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
```diff
@@ -45,12 +45,27 @@
...
+      for (int i = 0; i < dm.practitionerRole().get().location().size(); i++) {
```
jblefkove-va: this can be a for-each loop

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
```diff
@@ -45,12 +45,27 @@
...
+        if (dm.practitionerRole().get().location().get(i).type().isEmpty()) {
```
jblefkove-va: As Bryan mentioned, although it is unlikely to occur in practice here, collections may contain null elements. So you need to verify that `location().get(i)` is not null

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/DatamartPractitionerTransformerTest.java
```diff
@@ -0,0 +1,93 @@
...
+                .build());
+  }
+
+  @Test
+  public void nullChecks() {
+    assertThat(Dstu2PractitionerTransformer.healthcareServices(Optional.empty())).isNull();
+    assertThat(Dstu2PractitionerTransformer.roleCoding(null)).isNull();
+    assertThat(Dstu2PractitionerTransformer.role(null)).isNull();
+    assertThat(Dstu2PractitionerTransformer.telecom(null)).isNull();
+    // .isEqualTo("1990-12-12");
```
jblefkove-va: cruft

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
jjennings-va: It was not being used, (nor is it being used in LocationEntity.java) Do we need it?

jjennings-va: It was not being used (nor is it used in LocationEntity.java) Do we need it?

jblefkove-va: Since dstu2 practioner doesn't support any real searches, it is unused. So you can remove it... It can be restored later if necessary

jblefkove-va: Since dstu2 practioner doesn't support any real searches, it is unused. So you can remove it... It can be restored later if necessary

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+    return Address.builder()
+        .line(emptyToNull(Arrays.asList(address.line1(), address.line2(), address.line3())))
+        .city(address.city())
+        .state(address.state())
+        .postalCode(address.postalCode())
+        .build();
+  }
+
+  static String birthDate(Optional<LocalDate> source) {
+    LocalDate date = source.isEmpty() ? null : source.get();
```
bschofield-va: Optionals are fluent and functional... this entire method can be re-written as a single line.

```
return source.map(LocalDate::toString).orElse(null)
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+    if (date == null) {
+      return null;
+    }
+    return date.toString();
+  }
+
+  static Reference healthCareService(String s) {
```
bschofield-va: `s` makes me sad :(

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+        services.stream().map(s -> healthCareService(s)).collect(Collectors.toList()));
+  }
+
+  static HumanName name(DatamartPractitioner.Name source) {
+    if (source == null
+        || allBlank(source.family(), source.given(), source.prefix(), source.suffix())) {
+      return null;
+    }
+    return HumanName.builder()
+        .family(nameList(Optional.of(source.family())))
```
bschofield-va: If `source.family()` or `source.given()` are `null`, this will thrown an exception. You want to use `Optional.ofNullable(...)`, e.g. `Optional.ofNullable(source.family())`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+
+  static List<String> nameList(Optional<String> source) {
+    if (source.isEmpty() || source.get().isBlank()) {
+      return null;
+    }
+    return singletonList(source.get());
+  }
+
+  static CodeableConcept role(Optional<DatamartCoding> source) {
+    if (source == null || source.get().code().isEmpty() || source.get().code().get() == null) {
```
bschofield-va: The reason I'm so picky about using `Optional<X>` in our models is to reinforce the fact that their values may or may not be present. If you find yourself doing `someOptional.get()` without first checking to see if it's actually present, you're probably doing something wrong. 

The other thing is the is that `Optional` itself should never be `null`. Its value, maybe. In this case, `source` should never be `null`. It may be empty (`isEmpty()`), but not `null`.

One last though, `Optional` provides additional helper methods like `orElse` that can simplify the use. See the `birthDate` comment above.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+      return null;
+    }
+    return CodeableConcept.builder().coding(roleCoding(source.get())).build();
+  }
+
+  static List<Coding> roleCoding(DatamartCoding source) {
+    if (source == null || allBlank(source.system(), source.display(), source.code())) {
```
bschofield-va: This will fail with 
```
DatamartCoding.builder()
  .system(Optional.empty())
  .build()
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,194 @@
...
+      return null;
+    }
+    return convert(
+        source,
+        cdw ->
+            List.of(
+                Coding.builder()
+                    .code(cdw.code().get())
```
bschofield-va: `Optional.get()` go boom when empty

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerEntity.java
```diff
@@ -45,12 +46,27 @@
...
-    return Sort.by("cdwId").ascending();
-  }
-
   @SneakyThrows
   DatamartPractitioner asDatamartPractitioner() {
-    return JacksonConfig.createMapper().readValue(payload, DatamartPractitioner.class);
+    DatamartPractitioner dm =
+        JacksonConfig.createMapper().readValue(payload, DatamartPractitioner.class);
+
+    if (dm.practitionerRole().isPresent()
```
bschofield-va: _DRY_
If we update `DatamartReference` to have this:
```
  /**
   * Set the type if it's missing. Several datamart objects have missing types and this can be used
   * to correct the model.
   */
  public DatamartReference typeIfMissing(String type) {
    if (type().isEmpty()) {
      type(Optional.ofNullable(type));
    }
    return this;
  }
```
We can replace all this _if_ business with
```
    if (dm.practitionerRole().isPresent()) {
      dm.practitionerRole()
          .get()
          .managingOrganization()
          .ifPresent(mo -> mo.typeIfMissing("Organization"));
      dm.practitionerRole().get().location().forEach(loc -> loc.typeIfMissing("Location"));
    }
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -91,24 +87,26 @@ static HumanName name(DatamartPractitioner.Name source) {
...
+    if (source.isEmpty() || source.get() == null) {
       return null;
     }
     return CodeableConcept.builder().coding(roleCoding(source.get())).build();
   }
 
   static List<Coding> roleCoding(DatamartCoding source) {
-    if (source == null || allBlank(source.system(), source.display(), source.code())) {
+    if (source.code().isEmpty()
+        || source.code().get() == null
```
bschofield-va: `source.code().get() == null`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,192 @@
...
+    if (source.isEmpty() || source.get() == null) {
+      return null;
+    }
+    return CodeableConcept.builder().coding(roleCoding(source.get())).build();
+  }
+
+  static List<Coding> roleCoding(DatamartCoding source) {
+    if (source.code().isEmpty()
+        || source.code().get() == null
+        || allBlank(source.system(), source.display(), source.code())) {
```
bschofield-va: Can `source.sytem()`, `source.display()`, and `source.code()` ever be null ?
How does this affect `allBlank(...)`?
What does `DatamartCoding` have any utility methods that can help you?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,192 @@
...
+    if (source == null
+        || allBlank(
+            source.healthCareService(),
+            source.location(),
+            source.managingOrganization(),
+            source.role())) {
+      return null;
+    }
+    return Practitioner.PractitionerRole.builder()
+        .location(convert(source, cdw -> List.of(asReference(cdw.location().get(0)))))
```
bschofield-va: _NEVER_ index a list with out verifying it has a value at that location

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,192 @@
...
+    if (services.isEmpty()) {
+      return null;
+    }
+    return emptyToNull(
+        services.stream().map(s -> healthCareService(s)).collect(Collectors.toList()));
+  }
+
+  static HumanName name(DatamartPractitioner.Name source) {
+    if (source == null
+        || allBlank(source.family(), source.given(), source.prefix(), source.suffix())) {
```
bschofield-va: `prefix` and `suffix` are optional. `allBlank` won't work the way you expect.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  Practitioner.Gender gender(DatamartPractitioner.Gender source) {
+    return convert(
+        source, gender -> EnumSearcher.of(Practitioner.Gender.class).find(gender.toString()));
+  }
+
+  Practitioner.PractitionerRole practitionerRole(DatamartPractitioner.PractitionerRole source) {
+    if (source == null
+        || allBlank(
```
bschofield-va: won't work

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
jblefkove-va: Should we update `Transformers.isBlank()` (which is what `allBlank()` uses) to handle Optionals?

bschofield-va:
# REQUIRES: https://github.com/department-of-veterans-affairs/health-apis-deployer/pull/268

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,201 @@
...
+    return emptyToNull(
+        datamart.telecom().stream().map(tel -> telecom(tel)).collect(Collectors.toList()));
+  }
+
+  /** Convert the datamart structure to FHIR compliant structure. */
+  public Practitioner toFhir() {
+    return Practitioner.builder()
+        .id(datamart.cdwId())
+        .resourceType("Practitioner")
+        .active(datamart.active())
```
jblefkove-va: Potential null-pointer here: `datamart.active()` is `Boolean`, but this builder method accepts primitive `booelan`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,201 @@
...
+    return HumanName.builder()
+        .family(nameList(Optional.ofNullable(source.family())))
+        .given(nameList(Optional.ofNullable(source.given())))
+        .suffix(nameList(source.suffix()))
+        .prefix(nameList(source.prefix()))
+        .build();
+  }
+
+  static List<String> nameList(Optional<String> source) {
+    if (source.isEmpty() || source.get().isBlank()) {
```
jblefkove-va: This can be simplified to `Transformers.isBlank(source)` now

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,201 @@
...
+      return null;
+    }
+    return singletonList(source.get());
+  }
+
+  static CodeableConcept role(Optional<DatamartCoding> source) {
```
jblefkove-va: I think you can replace this method with a call to `Dstu2Transformers.asCodeableConceptWrapping()`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  static Reference healthCareService(String service) {
+    if (service.isBlank()) {
+      return null;
+    }
+    return Reference.builder().display(service).build();
+  }
+
+  static List<Reference> healthcareServices(Optional<String> services) {
```
jblefkove-va: parameter should be `service` singular

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,201 @@
...
+    if (services.isEmpty()) {
```
jblefkove-va: Now that `isBlank()` works with optionals, you can simplify this:
```
if (isBlank(service)) {return null;}
return List.of(Reference.builder().display(service.get()).build());
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,201 @@
...
+      return null;
+    }
+    return Practitioner.PractitionerRole.builder()
+        .location(
+            emptyToNull(
+                convert(
+                    source,
+                    dm ->
+                        List.of(
+                            dm.location().isEmpty() ? null : asReference(dm.location().get(0))))))
```
jblefkove-va: You still want to convert all the locations:
```
emptyToNull(source.location().stream().map(loc -> asReference(loc)).collect(Collectors.toList()))
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,171 @@
...
+    return Practitioner.PractitionerRole.builder()
+        .location(
+            emptyToNull(
+                source
+                    .location()
+                    .stream()
+                    .map(loc -> asReference(loc))
+                    .collect(Collectors.toList())))
+        .role(
+            Dstu2Transformers.asCodeableConceptWrapping(
```
jblefkove-va: Go ahead and statically import this method, since we already do that with the other transformers methods.

Also, there are several pairs of methods in `Dstu2Transformers` that are overloaded to accept either some object, or an optional for that object. (`asCoding()` for example) Let's add another `asCodeableConceptWrapping()` that accepts an optional to simplify this code here.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -0,0 +1,171 @@
...
+    return emptyToNull(
+        datamart.telecom().stream().map(tel -> telecom(tel)).collect(Collectors.toList()));
+  }
+
+  /** Convert the datamart structure to FHIR compliant structure. */
+  public Practitioner toFhir() {
+    return Practitioner.builder()
+        .id(datamart.cdwId())
+        .resourceType("Practitioner")
+        .active(datamart.active().booleanValue())
```
jblefkove-va: Unfortunately this is still a null pointer if `datamart.active()` is null. I think you want `BooleanUtils.isTrue(datamart.active())`

bschofield-va:
# DO NOT MERGE UNTIL PATHS WORK IS COMPLETE

PR #413, opened 2019-12-09
SSS-94 Add dstu2 prefix to utility classes

jblefkove-va:
https://vasdvp.atlassian.net/browse/SSS-94

To avoid a giant monster PR for supporting Stu3 location, this PR adds the `Dstu2` prefix to the `Bundler`, `Transformers`, and `Validator` classes.

`Stu3` implementations coming soon...

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 ### STS ###
 .apt_generated
+.checkstyle
 .classpath
 .factorypath
 .project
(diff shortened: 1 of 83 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@
 
 ### STS ###
 .apt_generated
+.checkstyle
```
jblefkove-va: This is the only non-renaming change

PR #416, opened 2019-12-09
SSS-86 Support STU-3 Location

jblefkove-va:
https://vasdvp.atlassian.net/browse/SSS-86

Add support for first STU3 resource, location. Supported searches described at https://www.fhir.org/guides/argonaut/pd/StructureDefinition-argo-location.html

New classes:
- `Stu3Bundler`
- `Stu3LocationController`
- `Stu3LocationTransformer`
- `Stu3Transformers`
- `Stu3Validator`

`ConfigurableBaseUrlPageLinks` and `MagicReferenceConfig` have been refactored to support STU3 alongside DSTU2

Note that `AbstractIncludesIcnMajig` is specific to DSTU2, but since locations are patient-agnostic, no changes are made there yet. The unused `LocationIncludesIcnMajig` class is removed.

This PR introduces new application properties, so https://github.com/department-of-veterans-affairs/health-apis-data-query-deployment/pull/47 should be merged first.

Next steps:
This PR does not include support for STU3 integration tests. I will be looking at those in a subsequent PR. This PR also does not make any changes to Swagger or the conformance statement. (I'm not sure what the approach there will be, or how the shared conformance application is impacted.)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/Stu3Transformers.java
```diff
@@ -0,0 +1,63 @@
...
+import gov.va.api.health.stu3.api.elements.Reference;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Optional;
+import lombok.experimental.UtilityClass;
+import org.apache.commons.lang3.StringUtils;
+
+/** Utility methods for STU3 transformers. */
+@UtilityClass
+public final class Stu3Transformers {
```
bschofield-va: Wonder if there should a generic `Transformers` with stuff like this, then version specific stuff in version specific transformer utilities?

jblefkove-va: 👍 I'll move the `allBlank` methods to a shared class

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/Stu3LocationController.java
```diff
@@ -0,0 +1,202 @@
...
+  @GetMapping
+  @SneakyThrows
+  public Location.Bundle searchByAddress(
+      @RequestParam(value = "address", required = false) String street,
+      @RequestParam(value = "address-city", required = false) String city,
+      @RequestParam(value = "address-state", required = false) String state,
+      @RequestParam(value = "address-postalcode", required = false) String postalCode,
+      @RequestParam(value = "page", defaultValue = "1") @Min(1) int page,
+      @CountParameter @Min(0) int count) {
+    if (street == null && city == null && state == null && postalCode == null) {
```
jblefkove-va: There may be a more elegant way to do this, but the intent is to support any combination of the address parameters, requiring at least one.

jblefkove-va:
This is ready to review

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/Stu3LocationController.java
bschofield-va: The exception and resulting message will be a little weird since it will complain about `address` specifically.
You could throw a `gov.va.api.health.dataquery.service.controller.ResourceExceptions.MissingSearchParameters` which will also be turned into a `400` Bad Request response. It may need a new constructor that allows you to craft a better error message.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/Stu3LocationController.java
```diff
@@ -0,0 +1,202 @@
...
+      throw new MissingServletRequestParameterException("address", "String");
+    }
+    Parameters pb = Parameters.builder().add("page", page).add("_count", count);
```
bschofield-va: I think we should update `Parameters` to allow `add("x",null)` or create `addUnlessNull(...)` so we don't need all this _if-ness_

data-query/src/test/java/gov/va/api/health/dataquery/tools/minimart/transformers/F2DDiagnosticReportTransformer.java
```diff
@@ -1,6 +1,6 @@
 package gov.va.api.health.dataquery.tools.minimart.transformers;
 
-import static org.codehaus.groovy.runtime.InvokerHelper.asList;
```
bschofield-va: nice catch

jblefkove-va: Eclipse caught it 😃

bschofield-va:
# REQUIRES: https://github.com/department-of-veterans-affairs/health-apis-deployer/pull/268

bschofield-va:
# REQUIRES: https://github.com/department-of-veterans-affairs/health-apis-deployer/pull/268

data-query/src/main/java/gov/va/api/health/dataquery/service/config/MagicReferenceConfig.java
```diff
@@ -56,19 +51,26 @@
...
   /** Property defining the references to serialize. */
   private final ReferenceSerializerProperties config;
 
   /** Auto-wired constructor. */
   @Autowired
   public MagicReferenceConfig(
       @Value("${argonaut.url}") String baseUrl,
-      @Value("${argonaut.base-path}") String basePath,
+      @Value("${argonaut.dstu2-base-path}") String dstu2BasePath,
+      @Value("${argonaut.stu3-base-path}") String stu3BasePath,
```
jhulbert-va: I think we should drop 'argonaut' from this.

jblefkove-va:
> There's already a lot going on here, so I think I'd save this for later: We should refactor the magic reference stuff to accept a _Strategy_ object that abstracts the details of working with specific versions. Then create Dstu2 and Stu3 strategy implementations so that the reference handling code doesn't need so much explicit _if dstu2_ and _if stu3_ and soon _if r4_ conditionals.
> 
> The search criteria isn't fully tested, but I suspect that's ok, we'll assume that Spring criteria is doing it's thing.
> 
> Questions:
> 
> * What's going to happen with Metadata?
> * When are integration tests coming?

I'm looking at integration tests next, but I don't want to clutter this PR with that. I'm less sure about the changes around Metadata (probably warrants a design discussion), but my guess is that we would rename `MetadataController` to `Dstu2MetadataController`, and create a new `Stu3MetadataController` alongside it just for the location resource.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/ConfigurableBaseUrlPageLinks.java
```diff
@@ -24,20 +22,37 @@
...
+  /** Base path for STU3 esources, e.g. api/stu3 */
+  private String stu3BasePath;
+
+  /** Spring constructor. */
   @Autowired
   public ConfigurableBaseUrlPageLinks(
-      @Value("${argonaut.url}") String baseUrl, @Value("${argonaut.base-path}") String basePath) {
+      @Value("${argonaut.url}") String baseUrl,
+      @Value("${argonaut.dstu2-base-path}") String dstu2BasePath,
+      @Value("${argonaut.stu3-base-path}") String stu3BasePath) {
```
jhulbert-va: Drop 'argonaut'

data-query/src/main/resources/application.properties
```diff
@@ -3,7 +3,8 @@ server.port=8090
 identityservice.url=unset
 mranderson.url=unset
 argonaut.url=unset
-argonaut.base-path=dstu2
+argonaut.dstu2-base-path=dstu2
+argonaut.stu3-base-path=stu3
```
jhulbert-va: Drop 'argonaut'

jblefkove-va: `stu3-base-path`?
`stu3-url-base-path`?
Something else?

jhulbert-va: I was thinking something like `dstu2.base-path` and `stu3.base-path`.

data-query/src/main/java/gov/va/api/health/dataquery/service/config/MagicReferenceConfig.java
bschofield-va: Yeah, dq is growing beyond just Argonaut, so this makes sense. We used a _public-url_/_public-path_ naming convention for Bulk to reinforce these properties are from the public consumer perspective.

```
data-query.public-url=https://api.va.gov
data-query.public-dstu2-base-path=services/fhir/v0/dstu2
data-query.public-stu3-base-path=services/fhir/v0/stu3
```

data-query/src/main/resources/application.properties
bschofield-va: See my comments on `application.properties`

jblefkove-va: image

bschofield-va:
# WAIT TO MERGE UNTIL PATHS WORK IS COMPLETE

PR #417, opened 2019-12-10
SSS-96 Separate generic transformers methods for STU3 use

jblefkove-va:
https://vasdvp.atlassian.net/browse/SSS-96

To reduce noise in https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/416, this PR moves the `allBlank` and `isBlank` methods out of `Dstu2Transformers`, so that they may be shared with STU3 transformers.


bschofield-va:
FYI @ilaflamme-va @dlennon-va @eclendenning-va @jhulbert-va

PR #423, opened 2019-12-12
SSS-98 STU3 location integration tests

jblefkove-va:
https://vasdvp.atlassian.net/browse/SSS-98

First STU3 integration tests. This does include any changes to the crawler.


data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -93,7 +95,8 @@ private static SystemDefinition local() {
     return SystemDefinition.builder()
         .ids(serviceDefinition("ids", url, 8089, null, "/api/"))
         .mrAnderson(serviceDefinition("mr-anderson", url, 8088, null, "/api/"))
-        .dataQuery(serviceDefinition("argonaut", url, 8090, null, "/dstu2/"))
+        .dstu2DataQuery(serviceDefinition("argonaut", url, 8090, null, "/dstu2/"))
+        .stu3DataQuery(serviceDefinition("argonaut", url, 8090, null, "/stu3/"))
```
jblefkove-va: This works locally, but for all the other environments, there is no distinction between `dstu2` and `stu3` in the api-path. I suspect this is a problem...

jblefkove-va:
I am going to hold off on addressing the merge conflicts here until https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/416 has merged

jblefkove-va:
This is ready to review

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/stu3/LocationIT.java
```diff
@@ -0,0 +1,35 @@
...
+import gov.va.api.health.stu3.api.resources.Location;
+import gov.va.api.health.stu3.api.resources.OperationOutcome;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+public class LocationIT {
+  ResourceVerifier verifier = ResourceVerifier.get();
+
+  @Category({Local.class
+    // , ProdDataQueryClinician.class
```
bschofield-va: We're going to want to turn these on in all environments, right?
Likewise, we'll want to turn them back on for DSTU2 and in the crawler deployment unit *.testvars. We turned them off when they broke with the stored procedure.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/stu3/ResourceVerifier.java
```diff
@@ -0,0 +1,134 @@
+package gov.va.api.health.dataquery.tests.stu3;
```
bschofield-va: stu3 and dstu2 versions of this class differ in minimal ways. I'm not in love with duplication. Once everything is datamarted, they will differ only by the path and test client instance. What would this look like if we merged these two and provided `ResourceVerifier.dstu2()` and `ResourceVerifier.stu3()` factories.

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
jblefkove-va: The `local` definition is the only one where the path has `dstu2` explicitly, so it's less obvious how to distinguish the `dstu2` and `stu3` paths for the others. For example prod and lab are `/services/argonaut/v0/`; does that become `/services/argonaut/v0/dstu2/` and `/services/argonaut/v0/stu3/`?

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/stu3/LocationIT.java
jblefkove-va: The resources are still broken in the stored procedure, correct? So to turn them back on, we should have them default to datamart and remove them from the cdw-and-datamart-resources list in `ResourceVerifier`?

bschofield-va: Resources still do not seem to work in stored proc. My concern is that we have a plan to get this turned on and don't leave it 99% done but not actually working in prod.

I spot checked a few Locations in QA, and they all came up Not Found for both Datamart and old crufty way. I've also reached out to the datasources team for clarification.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/AbstractResourceVerifier.java
```diff
@@ -4,125 +4,115 @@
...
 /** This support class can be used to test standard resource queries, such as reads and searches. */
 @Slf4j
-@NoArgsConstructor(access = AccessLevel.PRIVATE)
-public class ResourceVerifier {
-
-  private static final ResourceVerifier INSTANCE = new ResourceVerifier();
-
-  private static final String API_PATH = SystemDefinitions.systemDefinition().dataQuery().apiPath();
+@NoArgsConstructor(access = AccessLevel.PROTECTED)
+public abstract class AbstractResourceVerifier {
```
bschofield-va: I'm not mad at this, but it's not what I expected. Why the AbstractRV with Dstu2RV and Stu3RV implementations since the children do not alter the behavior?  I was assuming one configurable concrete implementation, e.g.

```
public static ResourceVerifier dstu2() {
  return ResourceVerifier.builder()
    .apiPath(...)
    .operationalOutcomeClass(...)
    .bundleClass(...)
    .testClient(...)
    .build()
}
```

Just curiuous

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/stu3/LocationIT.java
bschofield-va: I've created a separate story to capture this
https://vasdvp.atlassian.net/browse/SHANK-200

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/AbstractResourceVerifier.java
jblefkove-va: I fix! I fix!

PR #424, opened 2019-12-13
SHANK-178 Minimart database update performance improvements

dlennon-va:
WE ARE ZOOMING.

Full import runs in ~4 minutes vs up to an hour+ before


data-query/src/test/java/gov/va/api/health/dataquery/tools/minimart/MitreMinimartMaker.java
```diff
@@ -36,20 +36,30 @@
...
 import javax.persistence.EntityManager;
+import javax.persistence.EntityManagerFactory;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 
 @Slf4j
 public class MitreMinimartMaker {
 
+  private static ThreadLocal<EntityManager> localEntityManager = new ThreadLocal<>();
```
bschofield-va: `private static final ... LOCAL_ENTITY_MANAGER`

PR #428, opened 2019-12-16
Task/shank 164 refactor patient

ilaflamme-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/Dstu2PatientController.java
```diff
@@ -0,0 +1,252 @@
...
+    return Dstu2PatientTransformer.builder().datamart(dm).build().toFhir();
+  }
+
+  /** Return the raw Datamart document for the given identifier. */
+  @GetMapping(
+    value = "/{publicId}",
+    headers = {"raw=true"}
+  )
+  public String readRaw(@PathVariable("publicId") String publicId, HttpServletResponse response) {
+    PatientSearchEntity entity = readRaw(publicId);
```
jhulbert-va: Why not just use find by id here? The only thing that the `readRaw()` method does is return the result of `findById()`.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/Dstu2PatientController.java
```diff
@@ -0,0 +1,252 @@
...
+    AbstractIncludesIcnMajig.addHeader(response, entity.icn());
+    return entity.patient().payload();
+  }
+
+  PatientSearchEntity readRaw(String publicId) {
+    return findById(publicId);
+  }
```
jhulbert-va: If above comment is changed, this method can be deleted.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/Dstu2PatientController.java
ilaflamme-va: Woops. Cruft :/

PR #463, opened 2020-01-22
Task/shank 000 turn on its

jhulbert-va:


data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -244,20 +248,21 @@ private static Procedures productionCdwProcedures() {
...
+        .addressCity("ORLANDO")
+        .addressState("FL")
+        .addressPostalCode("32827")
+        .name("LAKE BALDWIN VAMC")
+        .npi("http://hl7.org/fhir/sid/us-npi|1558796409")
         .build();
   }
 
   private static TestIds.Practitioners productionPractitioners() {
     return TestIds.Practitioners.builder()
```
jblefkove-va: I would note in this method as well that these search criteria are tied to a real practitioner.

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -217,10 +217,14 @@ private static TestIds productionCdwIds() {
         .medicationStatement("I2-EIQB74V2APLMGKQJPRRT7LIPABT43MYPA2TEUW36N6BTEAJC65RA0000")
         .observation("I2-QSUC3WVCAOC7PWYON5HMETFYBQWCULOIQWLKHG6OP3DXH7M7MUTQ0000")
         .observations(observations())
-        .organization("I2-5R65QGBHXIK5FVLBMFXDK6CDNY000000")
+        // The following Organization is real (Orlando VAMC).
+        .organization("I2-U2YS4YSMVOGA4TNVOJB3RXVIQRQR7OXTDMFNC24L4YSKJKXSTCZA0000")
         .organizations(productionOrganizations())
         .patient("1011537977V693883")
-        .practitioner("I2-NCVNJO4N54ES6ZLYLYKKNFORRY000000")
+        // The following is a real Practitioner working at the Orlando VAMC. If test breaks, the
```
jblefkove-va: The code arranger sometimes garbles these comments that are trailing or embedded within a long chain of method calls, so it might be safer to make this a note at the beginning of the method.

PR #464, opened 2020-01-24
Shank 269 observation query perf

bschofield-va:
Alters behavior of `Observation.searchByPatientAndCategory` to select _all_ records and paginate in application code per Andrew Kelly's recommendation since SQL Server is performing a full table scan when paging criteria is part of the query.

This solution makes me sad.

It is enabled by default, but can be disabled per request with HTTP request header `query-hack:false`


bschofield-va:
jk

PR #466, opened 2020-01-27
Shank 268 update to parent 5

bschofield-va:
# REQUIRES department-of-veterans-affairs/health-apis-parent#49
`SNAPSHOT` will need to be removed before merging.

This PR contains no functionality changes


data-query/pom.xml
```diff
@@ -4,20 +4,21 @@
...
   <properties>
-    <fhir-resources.version>2.0.27</fhir-resources.version>
+    <fhir-resources.version>2.0.28-SNAPSHOT</fhir-resources.version>
     <guava.version>20.0</guava.version>
     <health-apis-ids.version>2.1.7</health-apis-ids.version>
     <jacoco.coverage>0.95</jacoco.coverage>
-    <swagger-maven-plugin.version>2.0.8</swagger-maven-plugin.version>
+    <swagger-maven-plugin.version>2.1.1</swagger-maven-plugin.version>
     <fall-risk.version>1.0.22</fall-risk.version>
     <mssql-jdbc.version>7.4.1.jre8</mssql-jdbc.version>
```
jblefkove-va: Should we go to `8.1.1.jre11`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/Dstu2PatientController.java
```diff
@@ -204,7 +205,9 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
             .name(name)
             .dates(birthdate)
             .build();
-    log.info("Looking for {} {}", name, spec);
+    URLEncoder.encode(name, StandardCharsets.UTF_8);
```
jblefkove-va: cruft?

data-query/pom.xml
bschofield-va: I kept with 7.4 because the MS SQL Server docs don't mention 8.1 as compatible with Sql Server 2017. They don't mention it at all. 
https://docs.microsoft.com/en-us/sql/connect/jdbc/microsoft-jdbc-driver-for-sql-server-support-matrix?view=sql-server-2017#sql-version-compatibility

PR #473, opened 2020-02-10
SHANK-301 Validate FHIR responses

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-301

Example messages:
```
WARN 21648 --- [nio-8090-exec-3] g.v.a.h.d.s.controller.ValidationAdvice  : read() response gov.va.api.health.stu3.api.resources.Practitioner failed validation: resourceType must not be blank, name.family must not be null, id must match "[A-Za-z0-9\-\.]{1,64}"

WARN 21648 --- [nio-8090-exec-2] g.v.a.h.d.s.controller.ValidationAdvice  : searchById(_id=[416704]) response gov.va.api.health.stu3.api.resources.Practitioner$Bundle failed validation: entry[0].resource.name.family must not be null, entry[0].resource.resourceType must not be blank, entry[0].resource.id must match "[A-Za-z0-9\-\.]{1,64}"
```

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/ValidationAdvice.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/ValidationAdvice.java
@@ -31,11 +31,7 @@ public final class ValidationAdvice implements ResponseBodyAdvice<Object> {
       ServletServerHttpRequest servletRequest = (ServletServerHttpRequest) request;
       sb.append("(")
           .append(
-              servletRequest
-                  .getServletRequest()
-                  .getParameterMap()
-                  .entrySet()
-                  .stream()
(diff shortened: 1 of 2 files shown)


jblefkove-va:
Ready

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/ValidationAdvice.java
```diff
@@ -38,7 +31,11 @@ public Object beforeBodyWrite(
       ServletServerHttpRequest servletRequest = (ServletServerHttpRequest) request;
       sb.append("(")
           .append(
-              servletRequest.getServletRequest().getParameterMap().entrySet().stream()
+              servletRequest
```
eclendenning-va: Looks good to me, but when i built, my formatter put 34-38 on the same line.

PR #475, opened 2020-02-12
SHANK-307 Upgrade to 6.0 parent with Error Prone

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-307

Drop SNAPSHOTs after https://github.com/department-of-veterans-affairs/health-apis-parent/pull/54

All code changes correspond to Error Prone findings

### DIFF
--- a/data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/crawler/SummarizingResultCollector.java
+++ b/data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/crawler/SummarizingResultCollector.java
@@ -135,7 +135,7 @@ public class SummarizingResultCollector implements ResultCollector {
   @Getter
   private static class SummaryStats {
     private int count;
-    private int totalTime;
+    private long totalTime;
     private int failures;
 
     void add(Result result) {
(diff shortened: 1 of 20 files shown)


data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/crawler/SummarizingResultCollector.java
```diff
@@ -135,7 +135,7 @@ private String resourceCountsSummary() {
   @Getter
   private static class SummaryStats {
     private int count;
-    private int totalTime;
+    private long totalTime;
```
jblefkove-va: Good job Error Prone!

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/DateTimeParameters.java
```diff
@@ -36,7 +36,7 @@ public DateTimeParameters(String paramString) {
     super();
     if (paramString.length() <= 1) {
       throw new IllegalArgumentException(
-          String.format("'{}' is not a valid date-time parameter", paramString));
+          String.format("'%s' is not a valid date-time parameter", paramString));
```
jblefkove-va: 🖐 🐛

PR #476, opened 2020-02-13
SHANK-317 Upgrade ID service

jblefkove-va:
https://vasdvp.atlassian.net/browse/SHANK-317

Depends on https://github.com/department-of-veterans-affairs/health-apis-ids/pull/16

### DIFF
--- a/data-query-ids-mapping/pom.xml
+++ b/data-query-ids-mapping/pom.xml
@@ -11,7 +11,7 @@
   <version>2.0.241-SNAPSHOT</version>
   <packaging>jar</packaging>
   <properties>
-    <health-apis-ids.version>2.1.10</health-apis-ids.version>
+    <health-apis-ids.version>3.0.0-SNAPSHOT</health-apis-ids.version>
     <jacoco.coverage>0.94</jacoco.coverage>
   </properties>
   <dependencies>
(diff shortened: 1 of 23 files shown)


data-query-ids-mapping/pom.xml
```diff
@@ -11,7 +11,7 @@
   <version>2.0.241-SNAPSHOT</version>
   <packaging>jar</packaging>
   <properties>
-    <health-apis-ids.version>2.1.10</health-apis-ids.version>
+    <health-apis-ids.version>3.0.0-SNAPSHOT</health-apis-ids.version>
```
ilaflamme-va: Drop snapshot after SHANK-316

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/WitnessProtectionTest.java
```diff
@@ -86,12 +86,13 @@ public void registerEmptyReturnsEmpty() {
...
     return Registration.builder()
         .uuid(id)
-        .resourceIdentity(
-            ResourceIdentity.builder()
-                .system("CDW")
-                .resource(resource)
-                .identifier(id + "cdw")
-                .build())
+        .resourceIdentities(
+            List.of(
```
ilaflamme-va: I already miss @Singular

PR #477, opened 2020-02-14
sun is getting real low big guy

eclendenning-va:
https://vasdvp.atlassian.net/browse/SHANK-321


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/PatientBulkFhirController.java
```diff
@@ -16,6 +17,7 @@
 import org.springframework.web.bind.annotation.RestController;
 
 @Validated
+@ConditionalOnProperty("bulk.fhir.on")
```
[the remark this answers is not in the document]
bschofield-va: Also, excellent solution.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/dstu2/PatientBulkFhirIT.java
```diff
@@ -104,4 +106,11 @@ public void bulkPatientCount() {
     var bulkFhirCount = response.expectValid(BulkFhirCount.class);
     assertThat(bulkFhirCount.count()).isGreaterThan(3);
   }
+
+  @Before()
+  public void isBulkFhirOn() throws IOException {
```
bschofield-va: The IT is skipped if the _default_ application properties defaults the bulk endpoint. But what if the _deployment_ configuration turns bulk endpoint on?

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/dstu2/PatientBulkFhirIT.java
```diff
@@ -23,19 +21,12 @@
...
 
   @Test
-  @Category(
-      value = {
-        Local.class,
-        LabDataQueryPatient.class,
-        ProdDataQueryPatient.class,
-        InternalApi.class
-      })
+  @Category(value = {LabBulkFhir.class, InternalApi.class})
```
jhulbert-va: If we add the `LabBulkFhir` category, we should get rid of the internal category here. What if we create another internal api and turn tests on for it?

PR #486, opened 2020-03-31
Update DQ for sandbox migration

ilaflamme-va:
also see : https://github.com/department-of-veterans-affairs/health-apis-data-query-deployment/pull/63


bschofield-va:
Will this be problematic?
```
$ git grep -l dev-api
data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/crawler/ResourceDiscoveryTest.java
data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/crawler/UrlReplacementRequestQueueTest.java
```

PR #489, opened 2020-05-05
Shank 596 r4 condition transformer

jhulbert-va:
https://vasdvp.atlassian.net/browse/SHANK-596


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionTransformer.java
```diff
@@ -0,0 +1,177 @@
...
+
+import gov.va.api.health.r4.api.datatypes.CodeableConcept;
+import gov.va.api.health.r4.api.datatypes.Coding;
+import gov.va.api.health.uscorer4.api.resources.Condition;
+import java.util.List;
+import lombok.Builder;
+
+@Builder
+public class R4ConditionTransformer {
+  private final DatamartCondition datamart;
```
jblefkove-va, line 15: `@NonNull` ?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionTransformer.java
```diff
@@ -0,0 +1,177 @@
...
+            .coding(
+                List.of(
+                    Coding.builder()
+                        .code("resolved")
+                        .display("Resolved")
+                        .system("http://hl7.org/fhir/R4/codesystem-condition-clinical.html")
+                        .build()))
+            .build();
+      default:
+        return null;
```
jblefkove-va, line 100: Should this be `IllegalArgumentException` as well?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionTransformer.java
```diff
@@ -0,0 +1,177 @@
...
+    return CodeableConcept.builder()
+        .text(snomedCode.display())
+        .coding(
+            List.of(
+                Coding.builder()
+                    .system("https://snomed.info/sct")
+                    .code(snomedCode.code())
+                    .display(snomedCode.display())
+                    .build()))
+        .text(snomedCode.display())
```
jblefkove-va, line 117: Should there be a check to verify that at least one of `snomedCode.display()` and `snomedCode.code()` is not blank?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionTransformer.java
```diff
@@ -0,0 +1,177 @@
...
+    return CodeableConcept.builder()
+        .text(icdCode.display())
+        .coding(
+            List.of(
+                Coding.builder()
+                    .system(systemOf(icdCode))
+                    .code(icdCode.code())
+                    .display(icdCode.display())
+                    .build()))
+        .text(icdCode.display())
```
jblefkove-va, line 134: Ditto, emptiness check?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionTransformer.java
```diff
@@ -0,0 +1,177 @@
...
+        .build();
+  }
+
+  private String systemOf(DatamartCondition.IcdCode icdCode) {
```
jblefkove-va, line 138: Let's add `@NonNull` on this parameter to make it more apparent that this will never be called with `null`.

PR #492, opened 2020-05-08
Rename Conformance to Metadata for use with R4

bschofield-va:
- Renames "Conformance" classes, fields, etc. to "Metadata" except for certain DSTU2 contexts where "Conformance" is the proper name.
- Changes application properties, too, which will require a DU change.

# Do not merge until DU is updated
### Afterwards, you can merge https://github.com/department-of-veterans-affairs/health-apis-data-query-deployment/pull/70


data-query/pom.xml
```diff
@@ -228,7 +228,7 @@
               <goal>timestamp-property</goal>
             </goals>
             <configuration>
-              <name>conformance.publication-date</name>
+              <name>metadata.publication-date</name>
```
jblefkove-va, line 231: There is also `conformance-timestamp` above

PR #502, opened 2020-05-12
Api 818 r4 openapi

jhulbert-va:
https://vajira.max.gov/browse/API-818

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DataQueryService.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DataQueryService.java
@@ -1,0 +1,58 @@
+package gov.va.api.health.dataquery.service.api;
+
+import gov.va.api.health.uscorer4.api.ConditionApi;
+import io.swagger.v3.oas.annotations.ExternalDocumentation;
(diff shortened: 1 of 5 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DataQueryService.java
```diff
@@ -0,0 +1,58 @@
...
+    security =
+        @SecurityRequirement(
+            name = "OauthFlow",
+            scopes = {"patient/Condition.read", "offline_access", "launch/patient"}),
+    info =
+        @Info(
+            title = "US Core",
+            version = "v1",
+            description =
+                " This service is compliant with the FHIR US Core Implementation"
```
jblefkove-va, line 27: Should we specify R4 here?

jhulbert-va: I went based on the title of the implementation guide. The like for argonaut liked to "Argonaut Data Query Implementation Guide" wheras the us-core ones title is just "US Core Implementation Guide" I could be convinced to add R4 though. It going to have to go through UX so they'll let us know if thats incorrect.

PR #507, opened 2020-05-17
r4 immunization transformer

ilaflamme-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
```diff
@@ -106,6 +106,8 @@ private void setEtlDate(String unused) {
     completed,
     @JsonProperty("entered-in-error")
     entered_in_error,
+    @JsonProperty("not-done")
```
jhulbert-va, line 109: Was this run by the datasources team?

bschofield-va: Yes, curious about adding a new value to the DM model

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationTransformer.java
```diff
@@ -0,0 +1,86 @@
...
+                    .build()))
+        .build();
+  }
+
+  Immunization toFhir() {
+    return Immunization.builder()
+        .resourceType(Immunization.class.getSimpleName())
+        .id(datamart.cdwId())
+        .status(status(datamart.status()))
+        // ._status() DAR in R4?
```
bschofield-va, line 76: I think DAR is appropriate in this case

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
ilaflamme-va: My thoughts were it was enumerated in R4, if it isn't doesn't exist we will just never use it

jhulbert-va: This should be modeled after the database in my opinion. If it can never be returned from the database, it shouldnt be here.

bschofield-va: I agree, the DM model is our contract with data sources team. We can't and shouldn't just change it with out discussion. It might be that they have enough information to determine this value, so we should bring it up with them.

Also, by adding this value, it forces a change to DSTU2 controller. What would we do there if `not_done` were to ever show up?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationTransformerTest.java
```diff
@@ -0,0 +1,70 @@
...
+                .build()
+                .toFhir())
+        .isEqualTo(Immunization.builder().resourceType("Immunization").build());
+  }
+
+  @Test
+  void note() {
+    assertThat(R4ImmunizationTransformer.note(Optional.empty())).isNull();
+    assertThat(R4ImmunizationTransformer.note(Optional.of("hello")))
+        .isEqualTo(ImmunizationSamples.R4.create().note("hello"));
```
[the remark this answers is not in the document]
ilaflamme-va: ```static List<Annotation> note(Optional<String> note) { ... } ``` 
Note returns a list of annotations and I thought this approach cleaned up the test.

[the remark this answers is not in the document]
jhulbert-va: Why not use the datamart samples then?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
ilaflamme-va: Removed for now, story in the backlog to address.

PR #509, opened 2020-05-18
API 235 r4 crawler part 2

bschofield-va:
- Crawler now supports R4
- r4 support UrlToResourceConverter.
- crawl-localhost support for r4

Sentinel test image does NOT invoke R4 crawler

This can be tested locally
```
$ PATIENT=40000200 TIME_LIMIT=PT30S ./crawl-localhost dstu2
$ PATIENT=40000200 TIME_LIMIT=PT30S ./crawl-localhost r4
```

Requires https://github.com/department-of-veterans-affairs/health-apis-data-query-deployment/pull/73


data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/crawler/UrlExtractor.java
```diff
@@ -0,0 +1,64 @@
...
+ * Process an object and extract URLs that can be followed while crawling. In general, this useful
+ * for processing bundle classes.
+ */
+public interface UrlExtractor {
+  /** Return a UrlExtractor that is suited for processing bundles of the given FHIR version. */
+  static UrlExtractor forFhirVersion(@NonNull FhirVersion fhirVersion) {
+    switch (fhirVersion) {
+      case DSTU2:
+        return new Dstu2BundleUrlExtractor();
+      case R4:
```
jblefkove-va, line 18: I would include a case for `STU3` here, to explicitly indicate that it is not supported. (vs. the more generic "unknown fhir version"0

PR #510, opened 2020-05-18
Api 836 r4 medication transformer

taylorrassmannlibertyits:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+import gov.va.api.health.r4.api.elements.Narrative;
+import gov.va.api.health.uscorer4.api.resources.Medication;
+import java.util.List;
+import java.util.Optional;
+import javax.validation.constraints.NotNull;
+import lombok.Builder;
+
+@Builder
+public class R4MedicationTransformer {
+  @NotNull final DatamartMedication datamart;
```
jblefkove-va, line 15: Did you mean `@NonNull`?

jblefkove-va, line 15: Should be private

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+
+  static CodeableConcept form(Optional<DatamartMedication.Product> maybeProduct) {
+    if (maybeProduct == null || !maybeProduct.isPresent()) {
+      return null;
+    }
+    DatamartMedication.Product product = maybeProduct.get();
+    return CodeableConcept.builder()
+        .text(product.formText())
+        .coding(List.of(Coding.builder().code(product.id()).display(product.formText()).build()))
```
jblefkove-va, line 24: Using the product ID twice (here, and as an ID for the medication itself) does not seem correct. At least for DSTU2, the `form` `CodeableConcept` only has the `text` field set anyway, so perhaps just remove this `coding`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+        datamart.rxnorm().isPresent() ? datamart.rxnorm().get().text() : datamart.localDrugName();
+    return Narrative.builder()
+        .div("<div>" + text + "</div>")
+        .status(Narrative.NarrativeStatus.additional)
+        .build();
+  }
+
+  List<Identifier> identifierFromProduct() {
+    if (datamart.product() != null && datamart.product().isPresent()) {
+      return List.of(Identifier.builder().id(datamart.product().get().id()).build());
```
jblefkove-va, line 82: Since `product` was removed from the spec, it's not 100% clear to me whether the datamart product ID should move up (as an ID for the medication as a whole), or down (as an ingredient ID), or if we should just omit it. (Complication, datamart does not have `ingredient.item`, which is a required field.) @bschofield-va ?

jhulbert-va: `productId` feels to be like it represents the product as a whole. Ingredient is a smaller part of that product. It doesnt feel to me like the `productId` should be in `ingredient`.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
taylorrassmannlibertyits: Ah I think the R4AllergyIntoleranceTransformer datamart needs to be private too. I see now the others are private.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Optional;
+import lombok.Builder;
+import lombok.NonNull;
+
+@Builder
+public class R4MedicationTransformer {
+  @NonNull private final DatamartMedication datamart;
+
+  static CodeableConcept form(Optional<DatamartMedication.Product> maybeProduct) {
+    if (maybeProduct == null || !maybeProduct.isPresent()) {
```
jhulbert-va, line 18: Optionals also have the concept of `isEmpty()` which will return the same boolean `! isPresent()` will

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
```diff
@@ -0,0 +1,98 @@
...
+  CodeableConcept bestCode() {
+    if (datamart.rxnorm() != null && datamart.rxnorm().isPresent()) {
+      DatamartMedication.RxNorm rxNorm = datamart.rxnorm().get();
+      return CodeableConcept.builder()
+          .coding(
+              List.of(
+                  Coding.builder()
+                      .code(rxNorm.code())
+                      .display(rxNorm.text())
+                      .system("https://www.nlm.nih.gov/research/umls/rxnorm")
```
jhulbert-va, line 50: should be 'http` not `https`

https://www.hl7.org/fhir/us/core/ValueSet-us-core-medication-codes.html#medication-clinical-drug-rxnorm

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medication/MedicationSamples.java
```diff
@@ -52,59 +44,72 @@ DatamartMedication medication(String cdwId) {
...
   public static class Dstu2 {
-
-    static Medication.Bundle asBundle(
-        String baseUrl, Collection<Medication> medications, int totalRecords, BundleLink... links) {
+    static gov.va.api.health.argonaut.api.resources.Medication.Bundle asBundle(
+        String baseUrl,
+        Collection<gov.va.api.health.argonaut.api.resources.Medication> medications,
+        int totalRecords,
+        gov.va.api.health.dstu2.api.bundle.BundleLink... links) {
       return Medication.Bundle.builder()
```
jhulbert-va, line 52: `gov.va.api.health.argonaut.api.resources.Medication.Bundle`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
bschofield-va:
```
if (someOptional == null)
if (someOptional != null)

```
It's bad form to check an `Optional` to be `null`.  Or more precisely, it's bad form for an `Optional` to ever possibly be `null`. This is sort of the whole point with them, is that the `Optional` itself will never be `null`. It may be empty, that is, its value is `null`, but the `Optional` itself will not.

Likewise, I never expect `List`, `Map`, `Set` or any other collection that we have control over to be `null`.
```
if (someList == null || someList.isEmpt()
```
makes me sad.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
taylorrassmannlibertyits: @bschofield-va ?

jblefkove-va: Based on Josh's comment, I'm okay with this current implementation

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
taylorrassmannlibertyits: Updated in next commit to not null check Optional.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
```diff
@@ -0,0 +1,95 @@
...
+   * <p>2. If rxnorm is not available, but product is, then this is a "local" drug. We will use the
+   * local drug name as the text and the code.coding.display value. The product.id is the VA
+   * specific "VUID" medication ID. We want to use this value as the code.coding.code along with the
+   * VA specific system.
+   *
+   * <p>3. If neither rxnorm or product is available, we'll create a code with no coding, using just
+   * the local drug name as text.
+   */
+  CodeableConcept bestCode() {
+    if (datamart.rxnorm() != null && datamart.rxnorm().isPresent()) {
```
jhulbert-va, line 39: These are optionals as well.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
```diff
@@ -36,7 +36,7 @@ static CodeableConcept form(Optional<DatamartMedication.Product> maybeProduct) {
    * the local drug name as text.
    */
   CodeableConcept bestCode() {
-    if (datamart.rxnorm() != null && datamart.rxnorm().isPresent()) {
+    if (!datamart.rxnorm().isEmpty()) {
```
jhulbert-va, line 39: `datamart.rxnorm().isPresent()`

PR #516, opened 2020-05-29
Api 966 r4 observation controller

jhulbert-va:
https://vajira.max.gov/browse/API-966


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
```diff
@@ -71,26 +71,44 @@ public Predicate toPredicate(
...
   @Value
-  class PatientAndCodesSpecification implements Specification<ObservationEntity> {
+  class PatientAndCodeAndDateSpecification implements Specification<ObservationEntity> {
     String patient;
 
     Set<String> codes;
 
+    DateTimeParameters date1;
+
+    DateTimeParameters date2;
```
taylorrassmannlibertyits, line 81: Does date1/date2 have more explanation behind them?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/R4ObservationController.java
```diff
@@ -0,0 +1,261 @@
...
+ */
+@Validated
+@RestController
+@RequestMapping(
+    value = {"/r4/Observation"},
+    produces = {"application/json", "application/fhir+json"})
+public class R4ObservationController {
+
+  // ToDo /$validate
+  // ToDo SHOULD patient, category, and status
```
taylorrassmannlibertyits, line 52: Leftover TODOs?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
bschofield-va: Let's do some better names here

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/R4ObservationController.java
jhulbert-va: I purposefully left them there. Those endpoint will need to be added, but can't be right now. I will pull them out into a task though.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
jhulbert-va: That's just the convention we use for these specs. Searches that allow dates can have between 0 and 2 dates. I can't think of a better way to name those.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/R4ObservationController.java
jhulbert-va: https://vajira.max.gov/browse/API-1015

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
taylorrassmannlibertyits: I'm ok with keeping date1/date2 after further talks about it.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/R4ObservationController.java
taylorrassmannlibertyits: Now if only I could see the link to verify ;)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
bschofield-va: Fair enough, I forget that the dates parameters include comparison criteria (GT2005-01-21)

PR #517, opened 2020-05-29
uscorer4 procedure transformer

ilaflamme-va:


data-query/pom.xml
```diff
@@ -12,7 +12,7 @@
   <packaging>jar</packaging>
   <properties>
     <fall-risk.version>1.0.29</fall-risk.version>
-    <fhir-resources.version>3.0.5</fhir-resources.version>
+    <fhir-resources.version>3.0.7-SNAPSHOT</fhir-resources.version>
```
taylorrassmannlibertyits, line 15: Remove SNAPSHOT once procedure model is merged

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/R4Transformers.java
```diff
@@ -16,6 +18,29 @@
...
+    }
+    return CodeableConcept.builder().coding(List.of(fhirCoding)).build();
+  }
+
+  /**
+   * Convert the optional coding to a FHIR coding and wrap it in a codeable concept. Returns null if
+   * it cannot be converted.
+   */
+  public static CodeableConcept asCodeableConceptWrapping(Optional<DatamartCoding> coding) {
+    if (coding == null || coding.isEmpty()) {
```
taylorrassmannlibertyits, line 38: You can just use isEmpty here.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/R4Transformers.java
```diff
@@ -16,6 +18,29 @@
...
+    }
+    return CodeableConcept.builder().coding(List.of(fhirCoding)).build();
+  }
+
+  /**
+   * Convert the optional coding to a FHIR coding and wrap it in a codeable concept. Returns null if
+   * it cannot be converted.
+   */
+  public static CodeableConcept asCodeableConceptWrapping(Optional<DatamartCoding> coding) {
+    if (coding.isEmpty()) {
```
jblefkove-va, line 38: `Transformers.isBlank()` here would also handle the case of an Optional that contains a blank string

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/R4ProcedureTransformer.java
```diff
@@ -0,0 +1,60 @@
...
+        return Status.not_done;
+      case completed:
+        return Status.completed;
+      default:
+        throw new IllegalArgumentException("Unsupported status: " + dmStatus);
+    }
+  }
+
+  static CodeableConcept statusReason(Optional<String> reasonNotPerformed) {
+    if (reasonNotPerformed.isEmpty()) {
```
bschofield-va, line 42: `isBlank( ... )` to capture the empty value

PR #518, opened 2020-05-29
Api 972 r4 procedure it tests

taylorrassmannlibertyits:
Tests for R4 Procedure Controller.

Once the controller exists:
- remove the "@Ignore" from the unit tests

- Uncomment the commented out code and remove the temporary objects

- Fix any potential misses


data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/ProcedureIT.java
```diff
@@ -0,0 +1,76 @@
...
+  @Delegate ResourceVerifier verifier = ResourceVerifier.r4();
+
+  @Test
+  @Ignore
+  @Category(Local.class)
+  public void advanced() {
+    // TODO: Uncomment line(s) below once R4 Procedure exists
+
+    //        verifier.verifyAll(
+    //                test(200, Procedure.Bundle.class, "Procedure?_id={id}",
```
jblefkove-va, line 25: Do these tests still need to be commented out since they are `@Ignore`d?

taylorrassmannlibertyits: For this test yes, since the verifier is the only part of the test and it requires R4 Procedure classes and they currently do not exist.

ilaflamme-va:
Branching off of this. Thanks.

PR #519, opened 2020-05-29
Fixed NPE in R4ObservationTransformer.

taylorrassmannlibertyits:


ilaflamme-va:
bad branch name. Jenkins wont build this.
`ABC-123-my-name`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/R4ObservationTransformer.java
```diff
@@ -276,7 +276,7 @@ static Quantity quantity(Optional<DatamartObservation.Quantity> maybeQuantity) {
       return null;
     }
     return Quantity.builder()
-        .value(BigDecimal.valueOf(quantity.value()))
+        .value((quantity.value() == null) ? null : BigDecimal.valueOf(quantity.value()))
```
jblefkove-va, line 279: pretty sure error-prone is going to complain about unnecessary parentheses

PR #528, opened 2020-06-11
Api 1030 medication request controller from medication order

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-1030


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/MedicationRequestFromMedicationOrderRepository.java
```diff
@@ -0,0 +1,16 @@
+package gov.va.api.health.dataquery.service.controller.medicationrequest;
```
bschofield-va, line 1: I think this should be deleted and just use the existing `MedicationOrderRepository`.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestController.java
```diff
@@ -0,0 +1,219 @@
...
+              .add("page", page)
+              .add("_count", count)
+              .build(),
+          count,
+          repository.findByIcn(
+              icn,
+              PageRequest.of(
+                  page - 1, count == 0 ? 1 : count, MedicationOrderEntity.naturalOrder())));
+    } else {
+      return MedicationRequest.Bundle.builder().build();
```
bschofield-va, line 209: too empty, it should have the paging links etc. just have empty `entry[]` field.

DR isn't exactly the same as the other resources, but the basic concept is here: https://github.com/department-of-veterans-affairs/health-apis-data-query/blob/master/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/Dstu2DiagnosticReportController.java#L249

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestControllerTest.java
```diff
@@ -0,0 +1,283 @@
...
+import org.mockito.Mockito;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.orm.jpa.DataJpaTest;
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@DataJpaTest
+@RunWith(SpringRunner.class)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+public class R4MedicationRequestControllerTest {
```
bschofield-va, line 38: Need a test for `intent=<something not order>` to assert that a valid, but empty bundle is returned.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/MedicationRequestFromMedicationOrderRepository.java
taylorrassmannlibertyits: Updated in next push.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestController.java
taylorrassmannlibertyits: It makes so much more sense now. Thanks for the example. Updated in next push.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
```diff
@@ -183,15 +185,24 @@ static Timing timing(
         .build();
   }
 
+  Extension requesterExtension(DatamartReference requester) {
```
ilaflamme-va, line 188: Can this guy get a unit test?

PR #529, opened 2020-06-11
Api 996 token searching prototype

eclendenning-va:
This is a prototype for token based searching detailed here : http://hl7.org/fhir/R4/search.html#token

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
@@ -36,7 +36,7 @@ public class QueryToken {
     }
     if (parameter.contains("|")) {
       return QueryToken.builder()
-          .system(parameter.substring(0, parameter.indexOf("|") - 1))
+          .system(parameter.substring(0, parameter.indexOf("|")))
           .code(parameter.substring((parameter.indexOf("|") + 1)))
           .mode(Mode.EXPLICIT_SYSTEM_EXPLICIT_CODE)
           .build();
@@ -72,7 +72,7 @@ public class QueryToken {
     return mode == Mode.NO_SYSTEM_EXPLICIT_CODE;
   }
 
-  private enum Mode {
+  public enum Mode {
     EXPLICIT_SYSTEM_EXPLICIT_CODE,
     EXPLICIT_SYSTEM_ANY_CODE,
     ANY_SYSTEM_EXPLICIT_CODE,
@@ -84,6 +84,10 @@ public class QueryToken {
       return new Behavior<T>().onAnySystemAndExplicitCode(action);
     }
 
+    public <T> Behavior<T> onExplicitSystemAndAnyCode(Function<String, T> action) {
+      return new Behavior<T>().onExplicitSystemAndAnyCode(action);
+    }
+
     public <T> Behavior<T> onExplicitSystemAndExplicitCode(BiFunction<String, String, T> action) {
       return new Behavior<T>().onExplicitSystemAndExplicitCode(action);
     }
@@ -92,13 +96,30 @@ public class QueryToken {
   public class Behavior<T> {
     private BiFunction<String, String, T> explicitSystemAndExplicitCode;
     private Function<String, T> anySystemAndExplicitCode;
+    private Function<String, T> explicitSystemAndAnyCode;
 
+    /** Execute correct behavior based on the mode of the token. */
+    @SneakyThrows
     public T execute() {
-      return explicitSystemAndExplicitCode.apply(code, system);
+      if (hasAnySystemAndExplicitCode()) {
+        return anySystemAndExplicitCode.apply(code);
+      }
+      if (hasExplicitSystemAndAnyCode()) {
+        return explicitSystemAndAnyCode.apply(system);
+      }
+      if (hasExplicitSystemAndExplicitCode()) {
+        return explicitSystemAndExplicitCode.apply(code, system);
+      }
+      throw new IllegalStateException("QueryToken in unsupported mode : " + mode);
     }
 
     public Behavior<T> onAnySystemAndExplicitCode(Function<String, T> action) {
       anySystemAndExplicitCode = action;
+      return this;
+    }
+
+    public Behavior<T> onExplicitSystemAndAnyCode(Function<String, T> action) {
+      explicitSystemAndAnyCode = action;
       return this;
     }
 
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
@@ -1,17 +1,20 @@
 package gov.va.api.health.dataquery.service.controller;
 
 import java.util.function.BiFunction;
+import java.util.function.Function;
 import lombok.Builder;
+import lombok.NonNull;
 import lombok.SneakyThrows;
-import org.springframework.data.jpa.domain.Specification;
+import lombok.Value;
 
 @Builder
+@Value
 public class QueryToken {
-  public String system;
+  String system;
 
-  public String code;
+  String code;
 
-  public Mode mode;
+  @NonNull Mode mode;
 
   /** Create a QueryToken from a token search parameter. */
   @SneakyThrows
@@ -41,33 +44,32 @@ public class QueryToken {
     return QueryToken.builder().code(parameter).mode(Mode.ANY_SYSTEM_EXPLICIT_CODE).build();
   }
 
-  public Behavior behavior() {
-    return new Behavior();
+  public BehaviorStemCell behavior() {
+    return new BehaviorStemCell();
   }
 
   public boolean hasAnySystemAndExplicitCode() {
-    return mode.equals(Mode.ANY_SYSTEM_EXPLICIT_CODE);
+    return mode == Mode.ANY_SYSTEM_EXPLICIT_CODE;
   }
 
   public boolean hasExplicitSystem() {
-    return mode.equals(Mode.EXPLICIT_SYSTEM_ANY_CODE)
-        || mode.equals(Mode.EXPLICIT_SYSTEM_EXPLICIT_CODE);
+    return mode == Mode.EXPLICIT_SYSTEM_ANY_CODE || mode == Mode.EXPLICIT_SYSTEM_EXPLICIT_CODE;
   }
 
   public boolean hasExplicitSystemAndAnyCode() {
-    return mode.equals(Mode.EXPLICIT_SYSTEM_ANY_CODE);
+    return mode == Mode.EXPLICIT_SYSTEM_ANY_CODE;
   }
 
   public boolean hasExplicitSystemAndExplicitCode() {
-    return mode.equals(Mode.EXPLICIT_SYSTEM_EXPLICIT_CODE);
+    return mode == Mode.EXPLICIT_SYSTEM_EXPLICIT_CODE;
   }
 
   public boolean hasNoSystem() {
-    return mode.equals(Mode.NO_SYSTEM_EXPLICIT_CODE);
+    return mode == Mode.NO_SYSTEM_EXPLICIT_CODE;
   }
 
   public boolean hasNoSystemAndExplicitCode() {
-    return mode.equals(Mode.NO_SYSTEM_EXPLICIT_CODE);
+    return mode == Mode.NO_SYSTEM_EXPLICIT_CODE;
   }
 
   private enum Mode {
@@ -77,16 +79,31 @@ public class QueryToken {
     NO_SYSTEM_EXPLICIT_CODE
   }
 
-  public class Behavior {
-    public BiFunction<String, String, Specification> explicitSystemAndExplicitCode;
+  public class BehaviorStemCell {
+    public <T> Behavior<T> onAnySystemAndExplicitCode(Function<String, T> action) {
+      return new Behavior<T>().onAnySystemAndExplicitCode(action);
+    }
 
-    public Specification execute() {
+    public <T> Behavior<T> onExplicitSystemAndExplicitCode(BiFunction<String, String, T> action) {
+      return new Behavior<T>().onExplicitSystemAndExplicitCode(action);
+    }
+  }
+
+  public class Behavior<T> {
+    private BiFunction<String, String, T> explicitSystemAndExplicitCode;
+    private Function<String, T> anySystemAndExplicitCode;
+
+    public T execute() {
       return explicitSystemAndExplicitCode.apply(code, system);
     }
 
-    public Behavior onExplicitSystemAndExplicitCode(
-        BiFunction<String, String, Specification> explicitSystemAndExplicitCode) {
-      this.explicitSystemAndExplicitCode = explicitSystemAndExplicitCode;
+    public Behavior<T> onAnySystemAndExplicitCode(Function<String, T> action) {
+      anySystemAndExplicitCode = action;
+      return this;
+    }
+
+    public Behavior<T> onExplicitSystemAndExplicitCode(BiFunction<String, String, T> action) {
+      explicitSystemAndExplicitCode = action;
       return this;
     }
   }
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
@@ -1,0 +1,93 @@
+package gov.va.api.health.dataquery.service.controller;
+
+import java.util.function.BiFunction;
+import lombok.Builder;
+import lombok.SneakyThrows;
+import org.springframework.data.jpa.domain.Specification;
+
+@Builder
+public class QueryToken {
+  public String system;
+
+  public String code;
+
+  public Mode mode;
+
+  /** Create a QueryToken from a token search parameter. */
+  @SneakyThrows
+  public static QueryToken parse(String parameter) {
+    if (parameter == null || parameter.isBlank() || parameter.equals("|")) {
+      throw new ResourceExceptions.BadSearchParameter(parameter);
+    }
+    if (parameter.startsWith("|")) {
+      return QueryToken.builder()
+          .code(parameter.substring(1))
+          .mode(Mode.NO_SYSTEM_EXPLICIT_CODE)
+          .build();
+    }
+    if (parameter.endsWith("|")) {
+      return QueryToken.builder()
+          .system(parameter.substring(0, parameter.length() - 1))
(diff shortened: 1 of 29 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
```diff
@@ -0,0 +1,131 @@
...
+  String system;
+
+  String code;
+
+  @NonNull Mode mode;
+
+  /** Create a QueryToken from a token search parameter. */
+  @SneakyThrows
+  public static QueryToken parse(String parameter) {
+    if (parameter == null || parameter.isBlank() || parameter.equals("|")) {
```
taylorrassmannlibertyits, line 22: isBlank does the null check for you.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
```diff
@@ -173,6 +189,46 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+      @RequestParam(value = "page", defaultValue = "1") @Min(1) int page,
+      @CountParameter @Min(0) int count) {
+    QueryToken statusToken = QueryToken.parse(status);
+    if (statusToken.hasExplicitSystem()) {
+      if (!statusToken.system().equals("http://hl7.org/fhir/event-status")) {
+        return Immunization.Bundle.builder().build();
+      }
+    }
+    if (statusToken.hasNoSystem()) {
+      return Immunization.Bundle.builder().build();
```
taylorrassmannlibertyits, line 206: since these return values are the same, can the logic be combined at all?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
```diff
@@ -7,6 +7,7 @@
 import gov.va.api.health.dataquery.service.controller.IncludesIcnMajig;
 import gov.va.api.health.dataquery.service.controller.PageLinks;
 import gov.va.api.health.dataquery.service.controller.Parameters;
+import gov.va.api.health.dataquery.service.controller.QueryToken;
```
taylorrassmannlibertyits, line 10: Unit tests for the added searchByPatientAndStatus?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
```diff
@@ -0,0 +1,131 @@
...
+public class QueryToken {
+  String system;
+
+  String code;
+
+  @NonNull Mode mode;
+
+  /** Create a QueryToken from a token search parameter. */
+  @SneakyThrows
+  public static QueryToken parse(String parameter) {
```
jblefkove-va, line 21: I don't necessarily have a better one right now, but `QueryToken` seems too generic a name

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public <T> Behavior<T> onExplicitSystemAndExplicitCode(BiFunction<String, String, T> action) {
+      return new Behavior<T>().onExplicitSystemAndExplicitCode(action);
+    }
+  }
+
+  public class Behavior<T> {
+    private BiFunction<String, String, T> explicitSystemAndExplicitCode;
+    private Function<String, T> anySystemAndExplicitCode;
+    private Function<String, T> explicitSystemAndAnyCode;
```
jblefkove-va, line 99: Should there also be a function for `noSystemAndExplicitCode`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
```diff
@@ -0,0 +1,131 @@
...
+    @SneakyThrows
+    public T execute() {
+      if (hasAnySystemAndExplicitCode()) {
+        return anySystemAndExplicitCode.apply(code);
+      }
+      if (hasExplicitSystemAndAnyCode()) {
+        return explicitSystemAndAnyCode.apply(system);
+      }
+      if (hasExplicitSystemAndExplicitCode()) {
+        return explicitSystemAndExplicitCode.apply(code, system);
```
jblefkove-va, line 111: arguments are swapped

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
eclendenning-va: `SystemCodeQueryToken` ?

jblefkove-va: Or `SystemAndCodeToken`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
```diff
@@ -0,0 +1,131 @@
...
+
+  public boolean hasAnySystemAndExplicitCode() {
+    return mode == Mode.ANY_SYSTEM_EXPLICIT_CODE;
+  }
+
+  public boolean hasExplicitSystem() {
+    return mode == Mode.EXPLICIT_SYSTEM_ANY_CODE || mode == Mode.EXPLICIT_SYSTEM_EXPLICIT_CODE;
+  }
+
+  public boolean hasExplicitSystemAndAnyCode() {
```
jblefkove-va, line 59: Some of these helper methods for the enum value seem unnecessary. If you have methods for each of the individual cases for system and code (`hasAnySystem(), hasExplicitSystem(), hasNoSystem(), hasAnyCode(), hasExplicitCode()`), you can combine them when necessary.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
```diff
@@ -0,0 +1,131 @@
...
+
+  public class Behavior<T> {
+    private BiFunction<String, String, T> explicitSystemAndExplicitCode;
+    private Function<String, T> anySystemAndExplicitCode;
+    private Function<String, T> explicitSystemAndAnyCode;
+
+    /** Execute correct behavior based on the mode of the token. */
+    @SneakyThrows
+    public T execute() {
+      if (hasAnySystemAndExplicitCode()) {
```
jblefkove-va, line 104: The body of this method seems like it should be a switch on `mode()`, particularly if there's a fourth case added for `noSystemAndExplicitCode`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/ImmunizationRepository.java
```diff
@@ -1,15 +1,81 @@
...
+      Predicate predicate = criteriaBuilder.equal(root.get("status"), code());
+      return criteriaBuilder.and(predicate);
+    }
+  }
+
+  @Value
+  class SystemSpecification implements Specification<ImmunizationEntity> {
+
+    @Builder
+    private SystemSpecification(String ignored) {}
```
jblefkove-va, line 49: Do we need the ignored parameter?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/ImmunizationRepository.java
```diff
@@ -1,15 +1,81 @@
...
+        CriteriaQuery<?> criteriaQuery,
+        CriteriaBuilder criteriaBuilder) {
+
+      Predicate predicate = criteriaBuilder.equal(root.get("status"), code());
+      return criteriaBuilder.and(predicate);
+    }
+  }
+
+  @Value
+  class SystemSpecification implements Specification<ImmunizationEntity> {
```
jblefkove-va, line 46: Should this be `AnySystemSpecification`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
```diff
@@ -173,6 +189,46 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+      }
+    }
+    if (statusToken.hasNoSystem()) {
+      return Immunization.Bundle.builder().build();
+    }
+    Specification<ImmunizationEntity> searchSpec =
+        statusToken
+            .behavior()
+            .onExplicitSystemAndExplicitCode(this::explicitSystemAndExplicitCode)
+            .onAnySystemAndExplicitCode(this::anySystemAndExplicitCode)
```
jblefkove-va, line 212: Can these functions be inlined? I think `.onAnySystemAndExplicitCode(c -> ImmunizationRepository.CodeSpecification.builder().code(c).build())` would be easier to understand.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
```diff
@@ -0,0 +1,131 @@
...
+      }
+      throw new IllegalStateException("QueryToken in unsupported mode : " + mode);
+    }
+
+    public Behavior<T> onAnySystemAndExplicitCode(Function<String, T> action) {
+      anySystemAndExplicitCode = action;
+      return this;
+    }
+
+    public Behavior<T> onExplicitSystemAndAnyCode(Function<String, T> action) {
```
jblefkove-va, line 121: This seems like a smell. Can these by lombok `@Builder` functions?

eclendenning-va: That was my inital thought but I got this :
https://github.com/rzwitserloot/lombok/issues/1302

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
eclendenning-va: These functions are going to be more complicated when systems determine which searchable column you're going to be using so I was doing this for consistency on other resources. Would it be better to inline them when I can?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
jblefkove-va: It can work with `Behavior` as a static class, you'd have to add the `queryToken` as a field. (And then QueryToken's `behavior` method would be something like `return Behavior.builder().queryToken(this)`)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
jblefkove-va: If they're simple inline them

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
eclendenning-va: I cant find a code that isnt bound to a system, so the specified null system case (noSystem)  is handled by the controller returning an empty bundle.

jblefkove-va: Is that only for Immunization though, or will that always be the case?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/ImmunizationRepository.java
eclendenning-va: Hmm should actually be `AnyCodeSpecification`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
eclendenning-va: Since entity cant really search by status, I think this is on hold.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/QueryToken.java
eclendenning-va: I left a comment on the story, it looked like that was always the case.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
```diff
@@ -173,6 +175,48 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+      @CountParameter @Min(0) int count) {
+    TokenParameter statusToken = TokenParameter.parse(status);
+    if ((statusToken.hasExplicitSystem()
+            && !statusToken.system().equals("http://hl7.org/fhir/event-status"))
+        || statusToken.hasNoSystem()) {
+      return Immunization.Bundle.builder().build();
+    }
+    Specification<ImmunizationEntity> searchSpec =
+        statusToken
+            .<Specification<ImmunizationEntity>>behavior()
```
bschofield-va, line 193: 🤮 
Use the StemCell pattern so you don't have to explicitly declare 
```
<Specification<ImmunizationEntity>>behavior()
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
```diff
@@ -173,6 +175,48 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
 
+  /** Search by patient and status. */
+  @GetMapping(params = {"patient", "status"})
+  public Immunization.Bundle searchByPatientAndStatus(
+      @RequestParam("patient") String patient,
+      @RequestParam("status") String status,
+      @RequestParam(value = "page", defaultValue = "1") @Min(1) int page,
+      @CountParameter @Min(0) int count) {
+    TokenParameter statusToken = TokenParameter.parse(status);
+    if ((statusToken.hasExplicitSystem()
```
bschofield-va, line 186: This seems like a common use case... how about add syntax sugar
```
token.hasSystem( "http://whatever" )
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,106 @@
...
+  public enum Mode {
+    EXPLICIT_SYSTEM_EXPLICIT_CODE,
+    EXPLICIT_SYSTEM_ANY_CODE,
+    ANY_SYSTEM_EXPLICIT_CODE,
+    NO_SYSTEM_EXPLICIT_CODE
+  }
+
+  @Builder
+  public static class Behavior<T> {
+    private BiFunction<String, String, T> onExplicitSystemAndExplicitCode;
```
bschofield-va, line 82: This is missing a state

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private Function<String, T> onAnySystemAndExplicitCode;
+
+    private Function<String, T> onExplicitSystemAndAnyCode;
+
+    private TokenParameter tokenParameter;
```
bschofield-va, line 88: drop this parameter and remove `static` modifier

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /** Execute correct behavior based on the mode of the token. */
+    @SneakyThrows
+    public T execute() {
+      switch (tokenParameter.mode) {
```
bschofield-va, line 93: missing a case

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
bschofield-va: This is better, but it binds to the token to a single behavior. The other way would multiple `behavior()` calls to support different return types.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,133 @@
...
+@Value
+@Builder
+public class TokenParameter {
+  String system;
+
+  String code;
+
+  @NonNull Mode mode;
+
+  /** Create a QueryToken from a token search parameter. */
```
ilaflamme-va, line 21: Create a _TokenParameter_ object from a token search parameter?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,133 @@
...
+  @SneakyThrows
+  public static TokenParameter parse(String parameter) {
+    if (isBlank(parameter) || parameter.equals("|")) {
+      throw new ResourceExceptions.BadSearchParameter(parameter);
+    }
+    if (parameter.startsWith("|")) {
```
ilaflamme-va, line 27: Just a thought, you could have used a switch here instead of chaining IF conditions. It might have looked a bit cleaner.

jblefkove-va: All the checks would have to be string equality for a switch

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
eclendenning-va: is this still applicable?

jblefkove-va: I think it should stay static to keep the Lombok builder

bschofield-va: fine, not going to fuss about that one, I care more about the end with the cleanest and easiest API to use

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
eclendenning-va: Since the real check would be  `if (hasExplicitUnsupportedSystem("http://whatever"))`I just added the positive individual case hasSupportedSystem("http://whatever") so it can be used in combination with other cases.

PR #532, opened 2020-06-18
Api 1271 post based integration tests

jhulbert-va:
[API-1271](https://vajira.max.gov/browse/API-1271)

### DIFF
--- a/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/R4PostSearchIT.java
+++ b/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/R4PostSearchIT.java
@@ -1,0 +1,59 @@
+package gov.va.api.health.dataquery.tests.r4;
+
+import gov.va.api.health.dataquery.tests.ResourceVerifier;
+import gov.va.api.health.dataquery.tests.categories.LabDataQueryPatient;
+import gov.va.api.health.dataquery.tests.categories.ProdDataQueryPatient;
+import gov.va.api.health.uscorer4.api.resources.AllergyIntolerance;
+import gov.va.api.health.uscorer4.api.resources.Condition;
+import gov.va.api.health.uscorer4.api.resources.Observation;
+import gov.va.api.health.uscorer4.api.resources.Patient;
+import java.util.Map;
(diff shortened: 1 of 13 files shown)


jhulbert-va:
Requires New Version of `health-apis-parent`: https://github.com/department-of-veterans-affairs/health-apis-parent/pull/62

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/R4PostSearchIT.java
```diff
@@ -0,0 +1,59 @@
...
+            verifier.ids().observations().dateRange().to()),
+        // Search by _id
+        test(
+            200,
+            Patient.Bundle.class,
+            "Patient/_search",
+            Map.of("Content-Type", "application/x-www-form-urlencoded"),
+            "_id={id}",
+            verifier.ids().patient()));
+  }
```
taylorrassmannlibertyits, line 58: Should there be a medication request search here too?

bschofield-va: See my comment yesterday. I don't think it's necessary since all of these requests are testing _kong_ not data query, and kong doesn't care what kind of resource it is.

jhulbert-va: Nope. We're trying to keep it with only a few. Realistically, it tests the kong plugin more that the application itself.

taylorrassmannlibertyits: Sounds good!

taylorrassmannlibertyits:
> I feel like this is too exhaustive for not much gain. That is, the tests aren't really testing functionality of data query, they are testing the Kong plugin... repeatedly. I think a focused "Post searching" test might be a better fit. It seems that _at most_ one request per resource, but I'm not convinced that much is necessary. Two or three combinations of body + request parameters is enough.

Aha, here it is. I jumped right into the code review.

PR #536, opened 2020-06-22
API-1232-ITs-exercise-fhir-v0

jhulbert-va:
[API-1232](https://vajira.max.gov/browse/API-1232)

### DIFF
--- a/data-query-tests/src/main/docker/entrypoint.sh
+++ b/data-query-tests/src/main/docker/entrypoint.sh
@@ -128,20 +128,28 @@ doCrawlerTest() {
   # If crawler test was specified and not explicitly told to skip then it's crawl time.
   if [ "$SKIP_CRAWLER" == "true" -o -z "$SENTINEL_CRAWLER" ]; then return; fi
 
+  setupForAutomation
+
   # Wipe out any included or excluded categories to make sure we pick up the crawler test class
   INCLUDE_CATEGORY=
   EXCLUDE_CATEGORY=
 
+  # This way, the ITs use defaults when INTERNAL_API_PATH isnt set,
+  # but the crawler will use the internal path always.
+  [ -z "$INTERNAL_API_PATH" ] && INTERNAL_API_PATH="/data-query"
+  DSTU2_CRAWL_PATH=${INTERNAL_API_PATH}${DSTU2_API_PATH:-/dstu2}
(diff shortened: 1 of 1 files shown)


data-query-tests/src/main/docker/entrypoint.sh
```diff
@@ -128,20 +128,28 @@ doCrawlerTest() {
   # If crawler test was specified and not explicitly told to skip then it's crawl time.
   if [ "$SKIP_CRAWLER" == "true" -o -z "$SENTINEL_CRAWLER" ]; then return; fi
 
+  setupForAutomation
+
   # Wipe out any included or excluded categories to make sure we pick up the crawler test class
   INCLUDE_CATEGORY=
   EXCLUDE_CATEGORY=
 
+  # This way, the ITs use defaults when INTERNAL_API_PATH isnt set,
```
taylorrassmannlibertyits, line 137: nitpick: missing apostrophe

data-query-tests/src/main/docker/entrypoint.sh
```diff
@@ -188,10 +191,10 @@ setupForAutomation() {
     addToSystemProperties "$property" "$DQ_URL"
   done
 
-  addToSystemProperties "sentinel.internal.api-path" "$INTERNAL_API_PATH"
-  addToSystemProperties "sentinel.dstu2.api-path" "$DSTU2_API_PATH"
-  addToSystemProperties "sentinel.stu3.api-path" "$STU3_API_PATH"
-  addToSystemProperties "sentinel.r4.api-path" "$R4_API_PATH"
+  [ -n "$INTERNAL_API_PATH" ] && addToSystemProperties "sentinel.internal.api-path" "$INTERNAL_API_PATH"
```
bschofield-va, line 194: I don't think this is quite right. If `DSTU2_API_PATH` is set, you'll product a command line that has the property defined twice with potentially different values.
```
INTERNAL_API_PATH=/ip
DSTU2_API_PATH=/d2
```
Will produce `java -Dsentinel.dstu2.api-path=/d2 ... -Dsentinel.dstu2.api-path=/ip/d2 ...`.

I do not think it is a good idea to define these properties in two different places, it's too easy to make a mistake. These should be removed below.

jhulbert-va: Sounds good. I'll look into it more.

PR #539, opened 2020-06-23
Api 1217 condition category token search

eclendenning-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,149 @@
...
+  public boolean hasSupportedCode(List<String> codes) {
+    return codes.stream().map(c -> hasSupportedCode(c)).anyMatch(b -> b == true);
+  }
+
+  public boolean hasSupportedSystem(String supportedSystem) {
+    return supportedSystem.equals(system);
+  }
+
+  public boolean hasSupportedSystem(List<String> systems) {
+    return systems.stream().map(s -> hasSupportedSystem(s)).anyMatch(b -> b == true);
```
jhulbert-va, line 90: Same, why not just `systems.stream().anyMatch(s -> hasSupportedSystem(s))`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -60,6 +62,17 @@ public R4ConditionController(
     this.witnessProtection = witnessProtection;
   }
 
+  private static Specification<ConditionEntity> explicitSystemSpec(String system) {
+    if ("http://terminology.hl7.org/CodeSystem/condition-category".equals(system)) {
+      return ConditionRepository.ExplicitSystemSpecification.builder()
+          .codes(List.of("problem", "diagnosis"))
```
jhulbert-va, line 68: Should we be using the enums toString() here instead of just strings?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -60,6 +62,17 @@ public R4ConditionController(
...
+          .build();
+    }
+    return ConditionRepository.ExplicitSystemSpecification.builder()
```
jhulbert-va, line 71: Should this be the default behavior? Doesnt seem right to assume `health-concern` if the category doesn't match the others.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -212,16 +223,48 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+            .build();
+    TokenParameter categoryToken = TokenParameter.parse(category);
+    if ((categoryToken.hasExplicitSystem()
+            && !categoryToken.hasSupportedSystem(
+                List.of(
+                    "http://hl7.org/fhir/R4/codesystem-condition-category.html",
+                    "http://terminology.hl7.org/CodeSystem/condition-category")))
+        || (categoryToken.hasExplicitCode()
+            && !categoryToken.hasSupportedCode(
+                List.of("problem-list-item", "encounter-diagnosis", "health-concern")))
```
jhulbert-va, line 241: Looking at those systems, is `health-concern` even supported anymore?

eclendenning-va: my mistake, the correct other system is http://hl7.org/fhir/us/core/CodeSystem/condition-category

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
eclendenning-va: true that doesnt seem quite right. Even though the only way to get here is if it is one of the two systems. I'll swap to two if statements and illegalstateexception to better illustrate that

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,149 @@
...
+
+  public boolean hasNoSystem() {
+    return mode == Mode.NO_SYSTEM_EXPLICIT_CODE;
+  }
+
+  public boolean hasSupportedCode(String supportedCode) {
+    return supportedCode.equals(code);
+  }
+
+  public boolean hasSupportedCode(List<String> codes) {
```
bschofield-va, line 81: You may consider changing this signature to support varargs instead of a list. It will clean up the usage a little.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
bschofield-va: While ok here, the `Enum.toString()` trick will fail to work properly anytime we use `@JsonProperty` annotations, e.g. Immunization's `in-progress` (the value being searched) vs.`in_progress` (the enum toString value)

bschofield-va, line 68: varargs would make this just a little nicer
```
.codes("problem","diagnosis")
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
bschofield-va: I feel like this pattern needs to be better encapsulated in Token. This is a 9 line if _condition_. The level of abstraction is too low and it's hard to understand _what_ the method is trying to do because it's buried in _how_ to do something. I'd rather see the _business rules_ isolated and sufficiently abstracted to express _intention_.

Perhaps something like this:

```
if ( !categoryToken.isSystemExplicitlySupported(supportedCategorySystems()) ) {
   return emptyBundle();
}
if (!categoryToken.isCodeExplicitlySupported(supportedCategoryCodes()) {
  return emptyBundle();
}
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionTransformer.java
```diff
@@ -44,7 +44,7 @@ CodeableConcept bestCode() {
                         Coding.builder()
                             .display("Encounter Diagnosis")
                             .code("encounter-diagnosis")
-                            .system("http://hl7.org/fhir/R4/codesystem-condition-category.html")
+                            .system("http://terminology.hl7.org/CodeSystem/condition-category")
```
bschofield-va, line 47: These systems should be made constants since we starting to use them in different places

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
```diff
@@ -1,20 +1,90 @@
...
   Page<ConditionEntity> findByIcnAndCategory(String icn, String category, Pageable pageable);
 
   Page<ConditionEntity> findByIcnAndClinicalStatusIn(
       String icn, Set<String> clinicalStatus, Pageable pageable);
+
+  @Value
+  class CodeSpecification implements Specification<ConditionEntity> {
+    String code;
+
+    @Builder
```
bschofield-va, line 37: Builders for objects with _one_ parameter can be made more simply with `@RequiredArgsConstructor(staticName = "of")`, changing this
```
CodeSpecification.builder().code("xxx").build()
```
(which can ofter get wrapped) 
into the more succinct 
```
CodeSpecification.of("xxx")
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -73,10 +86,8 @@ private String asDatamartCategory(String category) {
         return DatamartCondition.Category.problem.toString();
```
bschofield-va, line 86: Why not support the _enum_ directly, instead of Strings.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -212,16 +223,48 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+                List.of(
+                    "http://hl7.org/fhir/R4/codesystem-condition-category.html",
+                    "http://terminology.hl7.org/CodeSystem/condition-category")))
+        || (categoryToken.hasExplicitCode()
+            && !categoryToken.hasSupportedCode(
+                List.of("problem-list-item", "encounter-diagnosis", "health-concern")))
+        || categoryToken.hasNoSystem()) {
+      return bundle(parameters, emptyList(), 0);
+    }
+    Specification<ConditionEntity> searchSpec =
```
bschofield-va, line 245: This definitely will not yield the correct results since it drops _patient_ from the search criteria.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
bschofield-va: Also, I don't want to search by `code`. I want to search by `patient` **and** `code`. (also applied to the other specs)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
eclendenning-va: It does yield the correct results. The patient spec is `and`'ed with the category spec before being used by the repository. I think we can actually use this approach to cut some code duplication around our other specifications.

bschofield-va: I stand corrected! Nice!

bschofield-va: It does say something about the importance of the level of abstraction. The "category clause" is defined _twenty_ lines away from the patient clause. That's a kind of a long way. Some supporting methods could be useful to improve the LoA and provide clarity to the intention. Consider the following style... the business rules are more clearly visible and details on _how_ these rules are implemented are pushed to methods of a lower LoA.

```
    TokenParameter categoryToken = TokenParameter.parse(category);
     if (!categoryToken.isSystemExplicitlySupported(supportedCategorySystems())) {
       return emptyBundle();
    }
    if (!categoryToken.isCodeExplicitlySupported(supportedCategoryCodes()) {
      return emptyBundle();
    }
    String icn = witnessProtection.toCdwId(patient);
    return bundle(
        parameters,
        count,
        repository.findAll(
            PatientSpecification.of(icn).and(categoryClauseFor(categoryToken)),
            page(page, count)));
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -44,6 +46,9 @@
     value = {"/r4/Condition"},
     produces = {"application/json", "application/fhir+json"})
 public class R4ConditionController {
+  private final String problemAndDiagnosisSystem =
```
jhulbert-va, line 49: CAPS?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/TokenParameterTest.java
```diff
@@ -0,0 +1,163 @@
+package gov.va.api.health.dataquery.service.controller;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import org.junit.Test;
```
jhulbert-va, line 7: Junit5

jhulbert-va:
Should we add an integration tests or two for this?

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/ConditionIT.java
```diff
@@ -44,6 +44,16 @@ public void basic() {
             Condition.Bundle.class,
             "Condition?patient={patient}&category=health-concern",
             verifier.ids().patient()),
+        test(
+            200,
+            Condition.Bundle.class,
+            "Condition?patient={patient}&category=http://terminology.hl7.org/CodeSystem/condition-category|problem-list-item",
+            verifier.ids().patient()),
+        test(
```
jhulbert-va, line 52: What about `|[code]`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -65,18 +70,18 @@ public R4ConditionController(
...
-        return DatamartCondition.Category.problem.toString();
+        return DatamartCondition.Category.problem;
       case "encounter-diagnosis":
-        return DatamartCondition.Category.diagnosis.toString();
-      case "health-concern":
-        return "health-concern";
+        return DatamartCondition.Category.diagnosis;
       default:
-        throw new ResourceExceptions.BadSearchParameter("Invalid Category Parameter: " + category);
+        throw new IllegalStateException("Unsupported category code value: " + category);
```
bschofield-va, line 84: This isn't quite right. Question, what does the _user_ see if they search `category=foo`?
What should they see?

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/ConditionIT.java
jhulbert-va: I caught a bug while testing it manually that wasn't caught by the unit tests, but it would have been caught by the integration tests. I would agree they should be better sprinkled in there though as there are already multiple category ITs.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
eclendenning-va: The user would get an empty bundle back for searching `category=foo`. This case should never actually occur because all other values were handled by the `isCodeExplicitAndUnsupported` check.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/ConditionIT.java
bschofield-va: I've been pondering your response for a while ... and I'm curious. This seems like a failure of the unit test. Exactly what case was this?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,157 @@
...
+    private Function<String, T> onExplicitSystemAndAnyCode;
+
+    private BiFunction<String, String, T> onExplicitSystemAndExplicitCode;
+
+    private Function<String, T> onNoSystemAndExplicitCode;
+
+    /** Execute correct behavior based on the mode of the token. */
+    @SneakyThrows
+    public T execute() {
+      switch (token.mode) {
```
bschofield-va, line 124: observation ... we don't require that any particular handler is specified, but assume they can be used here, resulting in an NPE
```
.behavior()
.onAnySystemAndExplicitCode( f1 )
.onExplicitSystemAndAnyCode( f2 )
.execute() // NPE for EXPLICIT_SYSTEM_EXPLICIT_CODE or NO_SYSTEM_EXPLICIT_CODE
```
Perhaps we should check state here?
```
case ANY_SYSTEM_EXPLICIT_CODE:
   return check(onAnySystemAndExplicitCode).apply(toke.code);

...
<T> T check(T n) { 
  if (n == null){ throw new IllegalStateException("no handler specified for " + token.mode); }
  return n
}
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/TokenParameter.java
```diff
@@ -0,0 +1,165 @@
...
+    private Function<String, T> onAnySystemAndExplicitCode;
+
+    private Function<String, T> onExplicitSystemAndAnyCode;
+
+    private BiFunction<String, String, T> onExplicitSystemAndExplicitCode;
+
+    private Function<String, T> onNoSystemAndExplicitCode;
+
+    /** Check if behavior is specified before executing it. */
+    public <T1> T1 check(T1 n) {
```
bschofield-va, line 122: No change required, but `T1` is not the greatest the name. `F` for _function_  fits fantastically, friend!

PR #551, opened 2020-07-07
Api 1219 observation category token search

jhulbert-va:
[API-1219](https://vajira.max.gov/browse/API-1219)

### DIFF
--- a/data-query/spotbugs-excludes.xml
+++ b/data-query/spotbugs-excludes.xml
@@ -63,6 +63,7 @@
     <Or>
       <Class name="gov.va.api.health.dataquery.service.controller.observation.ObservationRepository$PatientAndCategoryAndDateSpecification$PatientAndCategoryAndDateSpecificationBuilder"/>
       <Class name="gov.va.api.health.dataquery.service.controller.observation.ObservationRepository$PatientAndCodeAndDateSpecification$PatientAndCodeAndDateSpecificationBuilder"/>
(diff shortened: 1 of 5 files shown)


data-query/spotbugs-excludes.xml
```diff
@@ -72,8 +73,21 @@
   </Match>
 
   <!--
-  These fields are going through parameter validation and will not have CRLF characters.
--->
+  This lombok generated builder accepts an array which is ultimated given to the PatientAndDateSpecification
```
eclendenning-va, line 76: ultimately*

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
```diff
@@ -29,43 +30,57 @@
...
+  @RequiredArgsConstructor(staticName = "of")
+  class CategorySpecification implements Specification<ObservationEntity> {
+    Set<String> categories;
+
+    @Override
+    public Predicate toPredicate(
+        Root<ObservationEntity> root,
+        CriteriaQuery<?> criteriaQuery,
+        CriteriaBuilder criteriaBuilder) {
+      List<Predicate> predicates = new ArrayList<>(1);
```
eclendenning-va, line 43: I dont think this array is buying you anything.
Try : `return criteriaBuilder.or(categoriesInClause);`

PR #552, opened 2020-07-08
Condition clinical-status token based searching

ilaflamme-va:
https://vajira.max.gov/browse/API-1218

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
@@ -59,10 +59,6 @@ public interface ConditionRepository
         CriteriaBuilder criteriaBuilder) {
       List<Predicate> predicates = new ArrayList<>(1);
       In<String> categoriesInClause = criteriaBuilder.in(root.get("clinicalStatus"));
-      /* Treat 'inactive' as resolved as inactive is the parent code of resolved */
-      if (clinicalStatuses.contains("inactive")) {
-        clinicalStatuses.add("resolved");
-      }
       clinicalStatuses.forEach(categoriesInClause::value);
       predicates.add(categoriesInClause);
       return criteriaBuilder.or(predicates.toArray(new Predicate[0]));
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaBuilder.In;
 import javax.persistence.criteria.CriteriaQuery;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
@@ -32,7 +33,7 @@ public interface ConditionRepository
 
   @RequiredArgsConstructor(staticName = "of")
   @Value
-  class CodeSpecification implements Specification<ConditionEntity> {
+  class CategoryCodeSpecification implements Specification<ConditionEntity> {
     DatamartCondition.Category code;
 
     @Override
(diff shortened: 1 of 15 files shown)


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionControllerTest.java
```diff
@@ -407,8 +426,158 @@ public void searchByPatientAndClinicalStatus() {
...
+                        10),
+                    link(
+                        BundleLink.LinkRelation.last,
+                        "http://fonzy.com/cool/Condition?clinical-status=resolved&patient=p1",
+                        1,
+                        10))));
+  }
+
+  @Test
+  public void searchByPatientAndClinicalStatusOfInactiveIsTreatedAtResolved() {
```
ilaflamme-va, line 459: Note, this part is a little weird.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
```diff
@@ -47,7 +48,30 @@ public Predicate toPredicate(
...
+    public Predicate toPredicate(
+        Root<ConditionEntity> root,
+        CriteriaQuery<?> criteriaQuery,
+        CriteriaBuilder criteriaBuilder) {
+      List<Predicate> predicates = new ArrayList<>(1);
+      In<String> categoriesInClause = criteriaBuilder.in(root.get("clinicalStatus"));
+      /* Treat 'inactive' as resolved as inactive is the parent code of resolved */
+      if (clinicalStatuses.contains("inactive")) {
+        clinicalStatuses.add("resolved");
+      }
```
jhulbert-va, line 65: Feels like this should be dealt with in the controller, not the repository.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionControllerTest.java
```diff
@@ -407,8 +426,158 @@ public void searchByPatientAndClinicalStatus() {
...
                         1,
                         10))));
+    assertThat(toJson(controller().searchByPatientAndClinicalStatus("p1", "resolved", 1, 10)))
+        .isEqualTo(
+            toJson(
+                ConditionSamples.R4.asBundle(
+                    "http://fonzy.com/cool",
+                    conditionsByPatient.get("p1").stream()
+                        .filter(c -> "Resolved".equalsIgnoreCase(c.clinicalStatus().text()))
+                        .collect(Collectors.toList()),
```
jhulbert-va, line 436: Why not move this up into a var and then use the list and `list.size()` here instead of using the same stream twice.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
ilaflamme-va: moved

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionRepository.java
```diff
@@ -47,7 +48,26 @@ public Predicate toPredicate(
...
+  class ClinicalStatusSpecification implements Specification<ConditionEntity> {
+
+    Set<String> clinicalStatuses;
+
+    @Override
+    public Predicate toPredicate(
+        Root<ConditionEntity> root,
+        CriteriaQuery<?> criteriaQuery,
+        CriteriaBuilder criteriaBuilder) {
+      List<Predicate> predicates = new ArrayList<>(1);
```
eclendenning-va, line 60: Dont need the array, and I think you mean `clinicalStatusesInClause`.

Try:

`return criteriaBuilder.or(clinicalStatusesInClause);`

ilaflamme-va: :+1:

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -122,22 +126,56 @@ public R4ConditionController(
...
+      List<TokenParameter> clinicalStatusTokens) {
+    Set<String> clinicalStatusesForQuery =
+        clinicalStatusTokens.stream()
+            .map(this::clinicalStatusFor)
+            .flatMap(Collection::stream)
+            .collect(Collectors.toSet());
+    /* Treat 'inactive' as resolved as inactive is the parent code of resolved */
+    if (clinicalStatusesForQuery.contains("inactive")) {
+      clinicalStatusesForQuery.add("resolved");
+    }
```
jhulbert-va, line 148:
```
Set<String> clinicalStatusesForQuery =
        clinicalStatusTokens.stream()
            .map(this::clinicalStatusFor)
            .flatMap(Collection::stream)
            .map(cs -> { if ("inactive".equals(cs)) { return "resolved"; } return cs; })
            .collect(Collectors.toSet());
```

jblefkove-va: `.map(this::clinicalStatusFor).flatMap(Collection::stream)` can be simplified to `flatMap(c -> clinicalStatusFor(c).stream())`

PR #553, opened 2020-07-09
Api 1220 observation code token based search

eclendenning-va:


data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/ObservationIT.java
```diff
@@ -61,9 +61,8 @@ public void basic() {
         test(
             200,
             Observation.Bundle.class,
-            "Observation?patient={patient}&code={loinc1}",
-            verifier.ids().patient(),
-            verifier.ids().observations().loinc1()),
+            "Observation?patient={patient}&code=http://loinc.org|",
```
jhulbert-va, line 64: Theres no search for one code IT after this change

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
```diff
@@ -152,4 +154,24 @@ public Predicate toPredicate(
...
+  class CodeSpecification implements Specification<ObservationEntity> {
+    Set<String> codes;
+
+    @Override
+    public Predicate toPredicate(
+        Root<ObservationEntity> root,
+        CriteriaQuery<?> criteriaQuery,
+        CriteriaBuilder criteriaBuilder) {
+      if (codes.contains(OBSERVATION_CODE_SYSTEM)) {
+        return criteriaBuilder.isNotNull(root.get("code"));
```
jhulbert-va, line 169: Should the others be updated? I think this is a better solution than a static list of values.

eclendenning-va: I'm going to go with no. If there are more than one system supported it doesn't work. I'm only using this because `loinc` has over 1000 codes which isnt plausible to have a list of.

PR #554, opened 2020-07-09
upgrade to java 14

eclendenning-va:


data-query-tests/pom.xml
```diff
@@ -4,11 +4,11 @@
...
-    <version>6.0.9</version>
+    <version>7.0.1</version>
     <relativePath/>
   </parent>
   <artifactId>data-query-tests</artifactId>
-  <version>2.0.314-SNAPSHOT</version>
+  <version>3.0.0-SNAPSHOT</version>
   <packaging>jar</packaging>
   <properties>
     <fhir-resources.version>3.0.15</fhir-resources.version>
```
taylorrassmannlibertyits, line 14: I think this should be 4.0.0?

data-query/pom.xml
```diff
@@ -4,11 +4,11 @@
...
-    <version>6.0.9</version>
+    <version>7.0.1</version>
     <relativePath/>
   </parent>
   <artifactId>data-query</artifactId>
-  <version>2.0.314-SNAPSHOT</version>
+  <version>3.0.0-SNAPSHOT</version>
   <packaging>jar</packaging>
   <properties>
     <fall-risk.version>1.0.29</fall-risk.version>
```
taylorrassmannlibertyits, line 14: Fhir resources version here too

PR #558, opened 2020-07-17
Api 1546 dstu2 diagnostic report v2

jhulbert-va:
[API-1546](https://vajira.max.gov/browse/API-1546)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/validation/DatamartValidationController.java
```diff
@@ -61,6 +63,9 @@ public Object validation(@RequestBody String payload) {
 
     JsonNode jsonNode = createMapper().readTree(payload);
     String objectType = jsonNode.get("objectType").asText();
+    if ("DiagnosticReport".equals(objectType)) {
```
bschofield-va, line 66: oh this is very gross. It's temporary, but very gross. I really want to make sure that whenever we cut support for V1, this if statement gets removed.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/Dstu2DiagnosticReportTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+import gov.va.api.health.dstu2.api.DataAbsentReason;
+import gov.va.api.health.dstu2.api.datatypes.CodeableConcept;
+import gov.va.api.health.dstu2.api.datatypes.Coding;
+import gov.va.api.health.dstu2.api.elements.Extension;
+import gov.va.api.health.dstu2.api.elements.Reference;
+import java.util.List;
+import java.util.stream.Collectors;
+import lombok.Builder;
+import lombok.NonNull;
+import org.springframework.util.CollectionUtils;
```
bschofield-va, line 17: ^ do be aware when creating dependencies on frameworks you do so very deliberately as it limits portability. For example, this transformer can not be used without Spring now, even though we're only using it to check if a collection is null or empty.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/Dstu2DiagnosticReportTransformer.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private Extension performerExtension() {
+    return datamart.accessionInstitution().isEmpty()
+        ? DataAbsentReason.of(DataAbsentReason.Reason.unknown)
+        : null;
+  }
+
+  private List<Reference> results() {
+    if (CollectionUtils.isEmpty(datamart.results())) {
```
bschofield-va, line 52: Our very own `Transformers.isBlank(...)` might have been a more portable choice.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/Dstu2DiagnosticReportController.java
```diff
@@ -69,29 +86,74 @@ public Dstu2DiagnosticReportController(
...
+  Pageable page(int page, int count) {
+    return PageRequest.of(page - 1, Math.max(count, 1), DiagnosticReportEntity.naturalOrder());
+  }
+
   /** Read by identifier. */
   @GetMapping(value = {"/{publicId}"})
   public DiagnosticReport read(
       @RequestHeader(name = "v2", defaultValue = "false") Boolean v2,
       @PathVariable("publicId") String publicId) {
     if (v2) {
```
bschofield-va, line 121: In general `if ( booleanObject )` is dangerous since `Boolean` variables have three states: `true`, `false`, and `null`. You're getting luck here because Spring is ensuring that `v2` is set thanks to `defaultValue` in the `@RequestHeader`. The safe check is `if (Boolean.TRUE.equals(v2))` since it is `null` safe.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/Dstu2DiagnosticReportController.java
```diff
@@ -275,7 +275,7 @@ void replaceReferences(Collection<DatamartDiagnosticReport> resources) {
       DiagnosticReportRepository.PatientAndCodeSpecification spec =
           DiagnosticReportRepository.PatientAndCodeSpecification.builder()
               .patient(cdwId)
-              .code(code)
+              .code("".equals(code) ? null : code)
```
bschofield-va, line 278: What should `code=" "` do?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/Dstu2DiagnosticReportControllerTest.java
```diff
@@ -356,29 +355,26 @@ void searchByPatientAndCategoryNoDate() {
...
                         BundleLink.LinkRelation.last,
-                        "http://fonzy.com/cool/DiagnosticReport" + "?category=LAB" + "&patient=p0",
+                        "http://fonzy.com/cool/DiagnosticReport?category=LAB&patient=p0",
                         1,
                         15))));
   }
 
   @Test
-  void searchByPatientAndCode() {
+  void searchByPatientAndCodeBlank() {
```
bschofield-va, line 374: This doesn't test for a _blank_ code... it tests an _empty_ code.

PR #560, opened 2020-07-22
Api 1573 r4 diagnoticreport transformer

jhulbert-va:
https://vajira.max.gov/browse/API-1573


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/R4DiagnosticReportTransformer.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.stream.Collectors;
+import lombok.Builder;
+import lombok.NonNull;
+
+@Builder
+public class R4DiagnosticReportTransformer {
+  @NonNull final DatamartDiagnosticReport datamart;
+
+  /** Determine the catgory value(s) for the DiagnosticReport. */
+  public List<CodeableConcept> category() {
```
bschofield-va, line 22: package protected maybe for all methods except `toFhir1?

PR #567, opened 2020-07-29
return empty bundle instead of error when searching with unknown system

eclendenning-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -135,12 +165,37 @@ public R4ConditionController(
         .execute();
   }
 
+  private boolean categoryIsUnsupported(TokenParameter t) {
```
jhulbert-va, line 168: Can we keep the syntax the same? Either `isSupported*`, `isUnupported*`, `*IsSupported`, or `*IsUnsupported`?

PR #575, opened 2020-08-19
API 1486 patient registration skeleton

bschofield-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/config/PatientRegistrationConfig.java
```diff
@@ -0,0 +1,22 @@
...
+
+@Configuration
+@ComponentScan(basePackages = {"gov.va.api.health.dataquery.patientregistration"})
+public class PatientRegistrationConfig {
+  @Bean
+  FilterRegistrationBean<PatientRegistrationFilter> patientRegistrationFilter(
+      @Autowired PatientRegistrar registrar) {
+    var registration = new FilterRegistrationBean<PatientRegistrationFilter>();
+    registration.setFilter(PatientRegistrationFilter.builder().registrar(registrar).build());
+    registration.addUrlPatterns("/dstu2/Patient/*", "/stu3/Patient/*", "/r4/Patient/*");
```
jhulbert-va, line 19: There is no `/stu3/Patient`

data-query-patient-registration/src/main/java/gov/va/api/health/dataquery/patientregistration/PatientRegistrationFilter.java
```diff
@@ -0,0 +1,108 @@
...
+    if (maybeRegistration == null) {
+      return;
+    }
+    if (!maybeRegistration.isDone()) {
+      log.warn("Patient registration is still in progress.");
+      return;
+    }
+    try {
+      PatientRegistration registration = maybeRegistration.get(10, TimeUnit.MILLISECONDS);
+      log.info("Patient registration has completed. {}", registration.access());
```
jhulbert-va, line 68: Should we log the whole registration? Otherwise, the ICN and registration information could be difficult to associate (they're different log messages).

data-query/src/main/java/gov/va/api/health/dataquery/service/config/PatientRegistrationConfig.java
bschofield-va: I know, but should there ever be I want this covered and not forgotten

data-query-patient-registration/src/main/java/gov/va/api/health/dataquery/patientregistration/PatientRegistrationFilter.java
bschofield-va: This is a temporary message that will change as we implement it the db portion

PR #576, opened 2020-08-19
r4 organization controller

ilaflamme-va:


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/organization/OrganizationSamples.java
```diff
@@ -4,15 +4,31 @@
...
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import lombok.AllArgsConstructor;
 
 public class OrganizationSamples {
 
+  // Organization Search Params
```
jhulbert-va, line 19: These should be used in `Dstu2` and `Stu3` as well to keep it uniform.

PR #585, opened 2020-08-27
Api 2186 extend patient registration filter first try baby

ilaflamme-va:


data-query-patient-registration/src/test/java/gov/va/api/health/dataquery/patientregistration/PatientRegistrationFilterSpringTest.java
```diff
@@ -33,10 +33,12 @@
   @Test
   void filterIsApplied() {
```
jhulbert-va, line 34: What about searchByIdentifier?

data-query-patient-registration/src/main/java/gov/va/api/health/dataquery/patientregistration/PatientRegistrationFilter.java
```diff
@@ -105,4 +105,22 @@ public String distillFromUri(HttpServletRequest request) {
       return null;
     }
   }
+
+  static class PatientSearchByIcnDistiller implements IcnDistiller {
+    private static final Pattern PATIENT_SEARCH_URI_PATTERN = Pattern.compile(".*/Patient.*");
+
+    @Override
+    public String distillFromUri(HttpServletRequest request) {
+      var matcher = PATIENT_SEARCH_URI_PATTERN.matcher(request.getRequestURI());
```
bschofield-va, line 114: I think this could be simplified with `request.getRequestURI().endsWith("/Patient")`
As is, this could fire on `/PatientBananna?_id=123`

ilaflamme-va: I've fixed this. I removed the REGEX, and am using this way now. There's a test to cover this case as well...

PR #592, opened 2020-09-21
refactor to use new model modules

eclendenning-va:


jblefkove-va:
don't forget `data-query-tests`

jblefkove-va (approved):

PR #595, opened 2020-09-23
API 1562: Creation of Controller for ETL Monitoring and Repository.

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,98 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@Slf4j
+@RestController
+@RequestMapping(
+    value = {"/etlstatus"},
+    produces = {"application/json"})
+public class LatestResourceEtlStatusController {
+  private final AtomicBoolean hasCachedCollectionStatus = new AtomicBoolean(false);
```
jblefkove-va, line 29: `collection-status` was a facilities-specific thing, those references should be updated

jblefkove-va:
This is going to need unit tests (https://github.com/department-of-veterans-affairs/lighthouse-facilities/blob/master/facilities/src/test/java/gov/va/api/lighthouse/facilities/HealthControllerTest.java) and integration tests (https://github.com/department-of-veterans-affairs/lighthouse-facilities/blob/master/facilities-tests/src/test/java/gov/va/api/lighthouse/facilities/tests/HealthIT.java)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,98 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@Slf4j
+@RestController
+@RequestMapping(
+    value = {"/etlstatus"},
```
jhulbert-va, line 26: Can we make this `/etl-status`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Gets health of etl resources.
+   *
+   * <p>Results are cached in collection-status to limit the amount of queries to once every 5
+   * minutes.
+   *
+   * <p>Every 5 minutes, the cache is invalidated.
+   */
+  @Cacheable("collection-status")
+  @GetMapping(value = "/general-status")
```
jhulbert-va, line 58: I think just `/etl-status` is fine for this endpoint. You can remove this `value` field.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,98 @@
...
+  public ResponseEntity<Health> collectionStatusHealth() {
+    hasCachedCollectionStatus.set(true);
+    Instant tooLongAgo = Instant.now().minus(1, ChronoUnit.DAYS);
+    AtomicBoolean overall = new AtomicBoolean(false);
```
jhulbert-va, line 62: I don't see this variable being used anywhere (it's being set, but not used).

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,98 @@
...
+public class LatestResourceEtlStatusController {
+  private final AtomicBoolean hasCachedCollectionStatus = new AtomicBoolean(false);
+
+  private LatestResourceEtlStatusRepository repository;
+
+  /** All args constructor. */
+  public LatestResourceEtlStatusController(
+      @Autowired LatestResourceEtlStatusRepository repository) {
+    this.repository = repository;
+  }
```
jhulbert-va, line 37: Instead of writing this out, you could use lombok:
```
@AllArgsConstructor(onConstructor_ = @Autowired)
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,98 @@
...
+            .withDetail("time", Instant.now())
+            .build();
+    if (!overallHealth.getStatus().equals(Status.UP)) {
+      return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).body(overallHealth);
+    }
+    return ResponseEntity.ok(overallHealth);
+  }
+
+  private Health toHealth(@NotNull LatestResourceEtlStatusEntity entity) {
+    Instant tooLongAgo = Instant.now().minus(1, ChronoUnit.DAYS);
```
jhulbert-va, line 91: I think we should pass `tooLongAgo` in from `collectionStatusHealth()` so that they're all being checked against the same instant.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,98 @@
...
+                  if (e.endDateTime().isBefore(tooLongAgo)) {
+                    overall.set(false);
+                  }
+                  return toHealth(e);
+                })
+            .collect(Collectors.toList());
+    Health overallHealth =
+        Health.status(
+                new Status(
+                    details.stream().anyMatch(h -> h.getStatus().equals(Status.DOWN))
```
bschofield-va, line 76: You computed `overall` just for this purpose

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private LatestResourceEtlStatusRepository repository;
+
+  /** Clears cache every 5 minutes. */
+  @Scheduled(cron = "0 */5 * * * *")
+  @CacheEvict(value = "resource-status")
+  public void clearResourceStatusScheduler() {
+    if (hasCachedResourceStatus.getAndSet(false)) {
+      // reduce log spam by only reporting cleared if we've actually cached something
+      log.info("Clearing etl-status cache");
```
jhulbert-va, line 40: Is it etl-status or resource-status? 
I think we should try to keep the same names when referring to the same thing.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,92 @@
...
+  @GetMapping(value = "/")
+  public ResponseEntity<Health> resourceStatusHealth() {
+    hasCachedResourceStatus.set(true);
+    Instant tooLongAgo = Instant.now().minus(1, ChronoUnit.DAYS);
+    AtomicBoolean overall = new AtomicBoolean(true);
+    Iterable<LatestResourceEtlStatusEntity> receivedData = repository.findAll();
+    if (StreamSupport.stream(receivedData.spliterator(), false).count() == 0) {
+      overall.set(false);
+    }
+    List<Health> details =
```
jhulbert-va, line 62: If it's empty, why are we streaming it again to get details?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusTest.java
```diff
@@ -0,0 +1,73 @@
...
+  private LatestResourceEtlStatusController _controller() {
+    return new LatestResourceEtlStatusController(repository);
+  }
+
+  private LatestResourceEtlStatusEntity _makeEntity(String resource, Instant time) {
+    return LatestResourceEtlStatusEntity.builder().resourceName(resource).endDateTime(time).build();
+  }
+
+  @Test
+  void resourceStatusHealth_clearCache() {
```
jhulbert-va, line 30: Repeating `resourceStatusHealth` on each test isn't really necessary. Instead name it after the method, or what its testing, e.g. `clearCache()` or `allResourcesHealthyIsOverallHealthy()`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,92 @@
...
+            .map(
+                e -> {
+                  if (e.endDateTime().isBefore(tooLongAgo)) {
+                    overall.set(false);
+                  }
+                  return toHealth(e, tooLongAgo);
+                })
+            .collect(Collectors.toList());
+    Health overallHealth =
+        Health.status(new Status(overall.get() == false ? "DOWN" : "UP", "Downstream services"))
```
bschofield-va, line 73: `== false` is a bit a of smell.
```
overall.get() == false ? "DOWN" : "UP
```

`overall` doesn't clearly indicate intent, because `overall` is named poorly. Both read fine, but one is wrong.
```
overall.get()?"DOWN":"UP"
overall.get()?"UP":"DOWN"
```

Perhaps you could indicate intent with a better name like, `allEndDatesWithinThreshold`?
```
allEndDatesWithinThreshold.get()?"UP":"DOWN"
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusRepository.java
```diff
@@ -0,0 +1,11 @@
...
+
+import gov.va.api.health.autoconfig.logging.Loggable;
+import org.springframework.data.repository.PagingAndSortingRepository;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Transactional;
+
+@Loggable
+@Transactional(isolation = Isolation.READ_UNCOMMITTED)
+public interface LatestResourceEtlStatusRepository
+    extends PagingAndSortingRepository<LatestResourceEtlStatusEntity, String> {}
```
bschofield-va, line 11: You don't need PagingAndSorting... find a smaller Repository interface to extend.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,92 @@
...
+   * <p>Results are cached in collection-status to limit the amount of queries to once every 5
+   * minutes.
+   *
+   * <p>Every 5 minutes, the cache is invalidated.
+   */
+  @Cacheable("resource-status")
+  @GetMapping(value = "/")
+  public ResponseEntity<Health> resourceStatusHealth() {
+    hasCachedResourceStatus.set(true);
+    Instant tooLongAgo = Instant.now().minus(1, ChronoUnit.DAYS);
```
bschofield-va, line 56: Do we think 24 hours is the right amount... I wonder if we should back off to 36. @jblefkove-va @jhulbert-va @eclendenning-va @ilaflamme-va

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,92 @@
...
+    if (hasCachedResourceStatus.getAndSet(false)) {
+      // reduce log spam by only reporting cleared if we've actually cached something
+      log.info("Clearing etl-status cache");
+    }
+  }
+
+  /**
+   * Gets health of etl resources.
+   *
+   * <p>Results are cached in collection-status to limit the amount of queries to once every 5
```
jblefkove-va, line 47: collection-status is copy-pasta

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
jblefkove-va: if the ETL doesn't run at exactly the same time every day (or takes longer on some days), there will be false positives

jhulbert-va: We've seen that ETL times can vary in the past. 36 hours sounds good.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,92 @@
...
+   * <p>Every 5 minutes, the cache is invalidated.
+   */
+  @Cacheable("resource-status")
+  @GetMapping(value = "/")
+  public ResponseEntity<Health> resourceStatusHealth() {
+    hasCachedResourceStatus.set(true);
+    Instant tooLongAgo = Instant.now().minus(1, ChronoUnit.DAYS);
+    AtomicBoolean overall = new AtomicBoolean(true);
+    Iterable<LatestResourceEtlStatusEntity> receivedData = repository.findAll();
+    if (StreamSupport.stream(receivedData.spliterator(), false).count() == 0) {
```
jblefkove-va, line 59: `Iterables.isEmpty(...)` ?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,92 @@
...
+      overall.set(false);
+    }
+    List<Health> details =
+        StreamSupport.stream(receivedData.spliterator(), true)
```
jblefkove-va, line 63: `Streams.stream(receivedData)`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,97 @@
...
+   *
+   * <p>Every 5 minutes, the cache is invalidated.
+   */
+  @Cacheable("resource-status")
+  @GetMapping(value = "/")
+  public ResponseEntity<Health> resourceStatusHealth() {
+    hasCachedResourceStatus.set(true);
+    Instant tooLongAgo = Instant.now().minus(36, ChronoUnit.HOURS);
+    AtomicBoolean areAllDatesAcceptable = new AtomicBoolean(true);
+    Iterable<LatestResourceEtlStatusEntity> receivedData = repository.findAll();
```
jblefkove-va, line 59: `receivedData` is a little too generic. How about `entities`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,97 @@
...
+                  })
+              .collect(Collectors.toList());
+    }
+
+    Health overallHealth =
+        Health.status(
+                new Status(areAllDatesAcceptable.get() ? "UP" : "DOWN", "Downstream services"))
+            .withDetail("name", "All downstream services")
+            .withDetail("downstreamServices", details != null ? details : "null")
+            .withDetail("time", Instant.now())
```
jblefkove-va, line 81: Since we also call `Instant.now()` earlier in the method, let's save that and reuse it here, so there is a single value for `now`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,97 @@
...
+                      areAllDatesAcceptable.set(false);
+                    }
+                    return toHealth(e, tooLongAgo);
+                  })
+              .collect(Collectors.toList());
+    }
+
+    Health overallHealth =
+        Health.status(
+                new Status(areAllDatesAcceptable.get() ? "UP" : "DOWN", "Downstream services"))
```
jblefkove-va, line 78: `Downstream services` is not really an accurate description for this. How about `Resource ETL statuses`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,97 @@
...
+            .withDetail("name", "All downstream services")
```
jblefkove-va, line 79: Since this just repeats the description above, this is redundant and can be removed.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,97 @@
...
+            .withDetail("downstreamServices", details != null ? details : "null")
```
jblefkove-va, line 80: How about `statuses` as this label

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,97 @@
...
+    List<Health> details = null;
+    if (Iterables.isEmpty(receivedData)) {
+      areAllDatesAcceptable.set(false);
+    } else {
+      details =
+          Streams.stream(receivedData)
+              .map(
+                  e -> {
+                    if (e.endDateTime().isBefore(tooLongAgo)) {
+                      areAllDatesAcceptable.set(false);
```
jblefkove-va, line 69: Modifying this variable as a side-effect while converting each database entity to a `Health` object is confusing. It would be clearer to calculate this afterward, separately, by doing another stream over the completed list of `Health` objects. This is what the corresponding Facilities health checks do. It is a little more complicated here since you also have to handle `details` being empty, so you would have something like:
```
boolean isDown = details.isEmpty() || details.stream().anyMatch(h -> h.getStatus().equals(Status.DOWN));
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,83 @@
...
+            .withDetail("time", now)
+            .withDetail("statuses", statusDetails.isEmpty() ? "null" : statusDetails)
+            .build();
+    if (!overallHealth.getStatus().equals(Status.UP)) {
+      return ResponseEntity.status(HttpStatus.SERVICE_UNAVAILABLE).body(overallHealth);
+    }
+    return ResponseEntity.ok(overallHealth);
+  }
+
+  private Health toHealth(LatestResourceEtlStatusEntity entity, Instant tooLongAgo) {
```
jblefkove-va, line 75: this method can be static

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,83 @@
...
+    Iterable<LatestResourceEtlStatusEntity> entities = repository.findAll();
+    List<Health> statusDetails =
+        Streams.stream(entities).map(e -> toHealth(e, tooLongAgo)).collect(Collectors.toList());
+    boolean isDown =
+        statusDetails.isEmpty()
+            || statusDetails.stream().anyMatch(h -> h.getStatus().equals(Status.DOWN));
+    Health overallHealth =
+        Health.status(new Status(isDown ? "DOWN" : "UP", "Resource ETL Statuses"))
+            .withDetail("time", now)
+            .withDetail("statuses", statusDetails.isEmpty() ? "null" : statusDetails)
```
jblefkove-va, line 67: Now that `statusDetails` is guaranteed to be non-null, is this check necessary? When it's empty it will show up in the response as `statusDetails : [ ]`, which is okay.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @Test
+  void allHealthyIsOverallHealthy() {
+    when(repository.findAll())
+        .thenReturn(
+            List.of(
+                _makeEntity("AllergyIntolerance", Instant.now()),
+                _makeEntity("Condition", Instant.now())));
+    ResponseEntity<Health> actual = _controller().resourceStatusHealth();
+    assertThat(actual.getBody().getStatus()).isEqualTo(Status.UP);
```
jblefkove-va, line 37: Can we update these tests to validate the entire response body? e.g.
```
    assertThat(actual.getBody())
        .isEqualTo(
            Health.status(new Status("UP", "Resource ETL Statuses"))
                .withDetail("time", now)
                .withDetail("statuses", List.of(...))
                .build());
```

(There is a complication that `Instant.now()` in your unit test is going to return a different value than `Instant.now()` in the controller. `Instant.now()` is a static method so you cannot mock/override it's behavior during a test. Is there another way to refactor to avoid this problem?)

jblefkove-va:
`curl -s http://localhost:8090/etl-status` returns 404

`curl -s http://localhost:8090/etl-status/` works as expected

jblefkove-va:
Caching is not enabled:
```
$ curl -s http://localhost:8090/etl-status/
{"description":"Resource ETL Statuses","status":"DOWN","details":{"time":"2020-09-25T12:47:55.412465200Z","statuses":"null"}}

$ curl -s http://localhost:8090/etl-status/
{"description":"Resource ETL Statuses","status":"DOWN","details":{"time":"2020-09-25T12:48:00.302328500Z","statuses":"null"}}

$ curl -s http://localhost:8090/etl-status/
{"description":"Resource ETL Statuses","status":"DOWN","details":{"time":"2020-09-25T12:48:01.193312500Z","statuses":"null"}}
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * Gets health of etl resources.
+   *
+   * <p>Results are cached in resource-status to limit the amount of queries to once every 5
+   * minutes.
+   *
+   * <p>Every 5 minutes, the cache is invalidated.
+   */
+  @Cacheable("resource-status")
+  @GetMapping(value = "/")
```
jhulbert-va, line 53: I think this is whats causing the behavior @jblefkove-va is talking about. The `@GetMapping` without the value should be enough for the controller to map requests here.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,90 @@
...
+  private LatestResourceEtlStatusRepository repository;
+
+  /**
+   * Compares Latest Resource ETL Status Entities to Health
+   *
+   * <p>Health is compared to 36 hours before it got called for validity.
+   */
+  private static Health toHealth(LatestResourceEtlStatusEntity entity) {
+
+    Instant tooLongAgo = Instant.now().minus(36, ChronoUnit.HOURS);
```
jblefkove-va, line 43: The previous version (passing in `tooLongAgo`) was preferable. This brings back the antipattern of invoking `Instant.now()` every time

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusTest.java
```diff
@@ -0,0 +1,110 @@
...
+@ExtendWith(MockitoExtension.class)
+public class LatestResourceEtlStatusTest {
+  @Mock LatestResourceEtlStatusRepository repository;
+
+  private LatestResourceEtlStatusController _controller() {
+    return new LatestResourceEtlStatusController(repository);
+  }
+
+  private Health _convertHealth(LatestResourceEtlStatusEntity entity, String expectedStatus) {
+    ;
```
jblefkove-va, line 26: ;

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Gets health of etl resources.
+   *
+   * <p>Results are cached in resource-status to limit the amount of queries to once every 5
+   * minutes.
+   *
+   * <p>Every 5 minutes, the cache is invalidated.
+   */
+  @Cacheable("resource-status")
+  @GetMapping()
```
jblefkove-va, line 70: unneccessary `()`

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusTest.java
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  @Test
+  void allHealthyIsOverallHealthy() {
+    Instant now = Instant.now();
+    LatestResourceEtlStatusEntity firstEntity = _makeEntity("AllergyIntolerance", now);
+    LatestResourceEtlStatusEntity secondEntity = _makeEntity("Condition", now);
+    when(repository.findAll()).thenReturn(List.of(firstEntity, secondEntity));
+    ResponseEntity<Health> actual = _controller().resourceStatusHealth();
+    Instant actualNow = (Instant) actual.getBody().getDetails().get("time");
```
jblefkove-va, line 43: Although it is less direct, this is still using a value generated by the code under test as part of the test expectation.

What I was getting at as a way to invoke `Instant.now()` just once for the whole test was to define a second method for testing that treats time as a parameter. The test invokes `Instant.now()` and passes that into the controller to set the time:
```
public ResponseEntity<Health> resourceStatusHealth() {
  return resourceStatusHealth(Instant.now());
}

ResponseEntity<Health> resourceStatusHealth(Instant now) {
  ...
}
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/etlstatus/LatestResourceEtlStatusController.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.cache.annotation.EnableCaching;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@Slf4j
+@EnableCaching
```
jblefkove-va, line 25: This should go on `Application.java`, in case we later want to use caching elsewhere in the app.

jblefkove-va (commented):
:shipit:

jhulbert-va (approved):

PR #597, opened 2020-09-28
API 1562 Created Integration Test

that-guy-is-it:


jblefkove-va (commented):
:shipit:

jblefkove-va:
Odd that `mitre.mv.db` actually decreased in size

jhulbert-va:
> Odd that `mitre.mv.db` actually decreased in size

I opened it, and I'm not seeing the app schema at all...

jhulbert-va (approved):

PR #599, opened 2020-10-01
Api 2397 observation null pointer exception

jjabando1:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/Dstu2Transformers.java
```diff
@@ -31,7 +31,11 @@ public static CodeableConcept asCodeableConceptWrapping(Optional<DatamartCoding>
     if (coding == null || coding.isEmpty()) {
       return null;
     }
-    return CodeableConcept.builder().coding(List.of(asCoding(coding.get()))).build();
+    Coding codeVal = asCoding(coding.get());
+    if (codeVal == null) {
+      return CodeableConcept.builder().coding(List.of()).build();
```
jhulbert-va, line 36: We never want to return an empty list to a consumer.

jjabando1: so return null instead?

jblefkove-va: Yes, return null

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/Dstu2Transformers.java
```diff
@@ -31,7 +31,11 @@ public static CodeableConcept asCodeableConceptWrapping(Optional<DatamartCoding>
     if (coding == null || coding.isEmpty()) {
       return null;
     }
-    return CodeableConcept.builder().coding(List.of(asCoding(coding.get()))).build();
+    Coding codeVal = asCoding(coding.get());
```
jblefkove-va, line 34: Is there a better name than `codeVal`? This function converts from datamart model to FHIR model; since the method parameter is already `coding`, how about `fhirCoding`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationTransformer.java
```diff
@@ -79,6 +79,8 @@ static CodeableConcept codeableConcept(Optional<DatamartObservation.CodeableConc
     Coding coding = asCoding(dmCode.coding());
     if (allBlank(coding, dmCode.text())) {
       return null;
+    } else if (isBlank(coding)) {
```
jblefkove-va, line 82: Additional branching is not preferred. There is a more elegant way to handle `coding` being null inline on line 85/87. Take a look at https://github.com/department-of-veterans-affairs/health-apis-data-query/blob/master/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/R4AllergyIntoleranceTransformer.java#L129

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationTransformer.java
```diff
@@ -96,6 +98,8 @@ static CodeableConcept codeableConcept(Optional<DatamartObservation.CodeableConc
     Quantity quantity = quantity(component.valueQuantity());
     if (allBlank(coding, quantity)) {
       return null;
+    } else if (isBlank(coding)) {
+      return Observation.ObservationComponent.builder().valueQuantity(quantity).build();
```
jblefkove-va, line 102: Instead of duplicating the builder, how about: `coding == null ? null : CodeableConcept.builder().coding(List.of(coding)).build()`

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/R4TransformersTest.java
```diff
@@ -40,6 +40,10 @@ public void asCodeableConceptWrappingReturnsValueIfCodingCanBeConverted() {
             CodeableConcept.builder()
                 .coding(List.of(Coding.builder().system("s").code("c").display("d").build()))
                 .build());
+    assertThat(
+            R4Transformers.asCodeableConceptWrapping(Optional.of(DatamartCoding.builder().build())))
+        .isEqualTo(
+            gov.va.api.health.r4.api.datatypes.CodeableConcept.builder().coding(List.of()).build());
```
jblefkove-va, line 46: Since this is now going to return `null`, move up into `asCodeableConceptWrappingReturnsNullIfCodingCannotBeConverted`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationTransformer.java
jjabando1: is that tertiary right? if coding == null, always return null? what if valueQuantity/quantity is not null? dont we want to return an ObservationComponent with just the valuequantity field populated?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/R4Transformers.java
```diff
@@ -36,7 +36,7 @@ public static CodeableConcept asCodeableConceptWrapping(DatamartCoding coding) {
    * it cannot be converted.
    */
   public static CodeableConcept asCodeableConceptWrapping(Optional<DatamartCoding> coding) {
-    if (Transformers.isBlank(coding)) {
+    if (Transformers.isBlank(coding) || asCoding(coding.get()) == null) {
```
jblefkove-va, line 39: Interesting that this check is slightly different from the dstu2/stu3 ones. While we're in here, boy scout rule: we usually prefer static imports of the `Transformers` methods

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationTransformer.java
```diff
@@ -98,10 +99,12 @@ static CodeableConcept codeableConcept(Optional<DatamartObservation.CodeableConc
       return null;
     }
 
-    return Observation.ObservationComponent.builder()
-        .code(CodeableConcept.builder().coding(List.of(coding)).build())
-        .valueQuantity(quantity)
-        .build();
+    return coding == null
```
jblefkove-va, line 102: Sorry for the confusion, I meant inlining the ternary as the argument to the `.code()` method:
```
    return Observation.ObservationComponent.builder()
        .code(coding == null ? null : CodeableConcept.builder().coding(List.of(coding)).build())
        .valueQuantity(quantity)
        .build();
```

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/Stu3TransformersTest.java
```diff
@@ -60,6 +60,10 @@ public void asReferenceReturnsNullWhenRefIsNull() {
   @Test
   public void codableConcept() {
     assertThat(asCodeableConceptWrapping(Optional.empty())).isNull();
+    assertThat(
+            Stu3Transformers.asCodeableConceptWrapping(
+                Optional.of(DatamartCoding.builder().build())))
+        .isNull();
```
jblefkove-va, line 66: For organization, move the other null test below, ` assertThat(asCodeableConceptWrapping(DatamartCoding.builder().build())).isNull();`, up with these two

jblefkove-va (commented):
:shipit:

jblefkove-va (approved):

PR #604, opened 2020-10-07
API-2657: R4PatientTransformer supports managingOrganization field

jjabando1:


jhulbert-va (changes requested):
Test?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
```diff
@@ -453,6 +452,14 @@ private String deceasedDateTime() {
     return emptyToNull(results);
   }
 
+  private Reference managingOrganization() {
+    Optional<String> managingOrganization = datamart.managingOrganization();
+    if (managingOrganization == null) {
```
jhulbert-va, line 457: `managingOrganization` is an `Optional<String>` there is a better way to check this.

jblefkove-va, line 457: null or empty

jhulbert-va: Where does `Datamart.create()` and `R4.create()` come from? Are they populating the managingOrganziation field?

jjabando1: Ah ok those are taken from PatientSamples. I think I need to update
public gov.va.api.health.r4.api.resources.Patient patient()'s returned Patient builder to include managingOrganization

jjabando1: So I assumed the better way to check was:
return managingOrganization.isPresent()
        ? Reference.builder().reference(managingOrganization.get()).build()
        : null;

This change causes most of the Controller and Transformer unit tests to fail. NullPointerExceptions for calls involving R4PatientTransformer.builder() and the controller(). I have no idea what is causing them to go null.

jblefkove-va: `Transformers.isBlank()` is the best way to verify that an Optional is both non-null and non-empty.

However, the fact that `datamart.managingOrganization()` is returning a null reference is also troubling. The different `DatamartXXX` model classes are supposed to include lazily-initializing getters for their `Optional` fields, e.g. https://github.com/department-of-veterans-affairs/health-apis-data-query/blob/master/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/DatamartAllergyIntolerance.java#L46

The fact that `DatamartPatient` is not doing this for `managingOrganization` is an oversight and should be corrected.

bschofield-va: Your reference is incomplete.

You would not be first person to ever want to convert a DatamartReference into an R4 Reference. This method can be exactly one line about 41 characters long.

Here's a hint
```
$ git grep DatamartReference data-query/src/main/java/ | grep R4
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
```diff
@@ -453,6 +452,14 @@ private String deceasedDateTime() {
     return emptyToNull(results);
   }
 
+  private Reference managingOrganization() {
```
bschofield-va, line 455: The big problem... you're adding production code, but you're not adding tests.

bschofield-va (dismissed):
how about tests?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
jjabando1: Should I go ahead and provide those lazy getters for the Optional fields in DatamartPatient?

jjabando1: In DatamartPatient, managingOrganization is of type Optional<String>, so I am not sure where I should be doing any kind of DatamartReference to R4Reference conversion. In creating my Reference object, should I just assign Reference's display field to the managingOrganization string, rather than the reference field?

jblefkove-va: Yes, add the lazy getters

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
bschofield-va: You haven't actually added any tests specifically for this.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
bschofield-va: My point about the the 41 character line to implement this method has been ignored.
See https://github.com/department-of-veterans-affairs/health-apis-data-query/blob/master/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/R4Transformers.java#L66

```
   return asReference(datamart.managingOrganization());
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
```diff
@@ -453,6 +455,20 @@ private String deceasedDateTime() {
...
```
bschofield-va, line 458: Changes this to be packaged protected and add explicit tests to verify the the couple of possible cases: null, partially populated, fully populated.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
```diff
@@ -539,6 +554,7 @@ Patient toFhir() {
         .deceasedDateTime(deceasedDateTime())
         .maritalStatus(maritalStatus())
         .contact(contacts())
+        .managingOrganization(managingOrganization(datamart.managingOrganization()))
```
jblefkove-va, line 557: It's more consistent for this method to remain non-static and access `datamart.managingOrganization()` internally. (In your unit test, you can set the `datamart` field on the transformer with the full test patient object and override just the `managingOrganization` field like so: `Datamart.create().patient().managingOrganization(...)`)

jblefkove-va (approved):

bschofield-va (changes requested):
hol up

bschofield-va (approved):

PR #606, opened 2020-10-09
API 2743 Patient Category Medication Requests

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
```diff
@@ -191,10 +197,25 @@ static Timing timing(
...
+    Pattern inPattern = Pattern.compile(patternInpatient);
+    Matcher matcherOut = outPattern.matcher(id);
+    Matcher matcherIn = inPattern.matcher(id);
+    String suffix = null;
+    if (matcherOut.matches()) {
+      suffix = "Outpatient";
+    } else if (matcherIn.matches()) {
+      suffix = "Inpatient";
+    }
+    return List.of(CodeableConcept.builder().text(suffix).build());
```
jhulbert-va, line 211: I think this mapping is missing pieces:
https://www.hl7.org/fhir/valueset-medicationrequest-category.html

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestControllerTest.java
```diff
@@ -74,7 +75,35 @@ R4MedicationRequestController controller() {
...
+  }
+
+  @Test
+  public void correctSuffixDetermination() {
+    var fhir = MedicationRequestSamples.R4.create();
+    MedicationRequest req = fhir.medicationRequestFromMedicationOrder();
+    assertThat(req).isNotNull();
+
+    req = fhir.medicationRequestFromMedicationOrder("123:");
+    assertThat(req.category()).isEqualTo(List.of(CodeableConcept.builder().text(null).build()));
```
jhulbert-va, line 90: This isn't right. If the id can't be parsed, the category should be null, not a list of empty CC's.

data-query/src/main/resources/application.properties
```diff
@@ -70,3 +70,5 @@ count-parameter.max-count=100
 bulk-fhir.enabled=false
 bulk.patient.maxRecordsPerPage=5000
 
+pattern.inpatient=.*:(I|FPI)
+pattern.outpatient=.*:(O|FP)
```
jhulbert-va, line 74: I think this would be FPO to match FPI above.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
that-guy-is-it: I am not sure I understand what is missing. The coding? CodeableConcept has only `text` and `coding`, according to:
https://www.hl7.org/fhir/datatypes.html#CodeableConcept
Or are you referring to the `Discharge` and `Community` codes? I was under the impression that only `Inpatient` and `Outpatient` were relevant to this story.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestController.java
```diff
@@ -61,16 +61,24 @@
...
-      @Autowired WitnessProtection witnessProtection) {
+      @Autowired WitnessProtection witnessProtection,
+      @Value("${pattern.outpatient}") String patternOutpatient,
+      @Value("${pattern.inpatient}") String patternInpatient) {
     this.bundler = bundler;
     this.medicationOrderRepository = medicationOrderRepository;
     this.medicationStatementRepository = medicationStatementRepository;
     this.witnessProtection = witnessProtection;
+    this.patternOutpatient = patternOutpatient;
+    this.patternInpatient = patternInpatient;
```
jblefkove-va, line 81: Since the patterns don't change, perhaps the patterns should be compiled once here, vs. compiling the patterns in each instance of the transformer?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
```diff
@@ -70,6 +72,10 @@
 
   @NonNull final DatamartMedicationOrder datamart;
 
+  private final String patternInpatient;
+
+  private final String patternOutpatient;
```
jblefkove-va, line 77: are these also `@NonNull`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestController.java
bschofield-va: +1

data-query/src/main/resources/application.properties
bschofield-va: -1
it should not, unless you want to go change all of the IDs for Frankenpatient.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
```diff
@@ -191,10 +197,25 @@ static Timing timing(
         .build();
   }
 
+  private List<CodeableConcept> parseCategory(String id) {
+    Pattern outPattern = Pattern.compile(patternOutpatient);
+    Pattern inPattern = Pattern.compile(patternInpatient);
+    Matcher matcherOut = outPattern.matcher(id);
+    Matcher matcherIn = inPattern.matcher(id);
+    String suffix = null;
```
bschofield-va, line 205: this value isn't a _suffix_. You need a better name.

data-query/src/main/resources/application.properties
jhulbert-va: As it stands currently, all Medication orders in FP have the suffix `:FP` feels like they should be edited so half are outpatient and half are inpatient?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
bschofield-va: To Peter's comment, you should pre-compile the pattern and store that vs. compiling each time. More over, this still has a some design issue that are worth noting to learn from.

You're paying the expense of building the _both_ pattern objects and _both_ matcher objects even though you are likely to only need one.

Additionally, this can emit a completely useless CodeableConcept if the ID matches neither.

I'd expect the method to look something like this. Note that it's package protected so you an write some tests for it.

```
List<CodeableConcept> categoryForId(String id) {
  if (inpatientPattern.matcher(id).matches() {
     return List.of(CodeableConcept.builder()...
  }
  if (outpatientPattern.matcher(id).matches() {
     return List.of(CodeableConcept.builder()...;
  }
  return null;
}
```

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestControllerTest.java
bschofield-va: I'd bet you wrote the code first, then the test. Try doing it in the opposite order. Write the test to product the result you _want_, then that will guide the behavior of the implementation. When you write the code first, you inevitably create test bias and the test reflects whatever behavior you implemented.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformerTest.java
```diff
@@ -112,6 +112,8 @@ public void toFhir() {
     assertThat(
             R4MedicationRequestFromMedicationOrderTransformer.builder()
                 .datamart(dmo)
+                .patternOutpatient(".*:(O|FP)")
```
bschofield-va, line 115: Not sure I'm a fan of the backwards naming. You might notice that we chose names that help _reading_... an attempt to approximate natural language. I understand what you've done, but not sure I love it.

How do other's feel about `outpatientPattern` vs. `patternOutpatient`?

jblefkove-va: I like `pattern.outpatient` for the property but `outpatientPattern` for the variable

jhulbert-va: :arrow_up:

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
```diff
@@ -70,6 +72,10 @@
 
   @NonNull final DatamartMedicationOrder datamart;
 
+  private final Pattern outPattern;
```
jblefkove-va, line 75: Should these also be `@NonNull`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
```diff
@@ -191,10 +197,36 @@ static Timing timing(
         .build();
   }
 
+  private List<CodeableConcept> parseCategory(String id) {
+    String system = "terminology.hl7.org/CodeSystem/medicationrequest-category";
+    if (outPattern.matcher(id).matches()) {
+      String displayText = "Outpatient";
+      String code = "outpatient";
```
jblefkove-va, line 204: If this is only used in one place, it's probably more readable to inline it

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestFromMedicationOrderTransformer.java
```diff
@@ -191,10 +197,36 @@ static Timing timing(
         .build();
   }
 
+  private List<CodeableConcept> parseCategory(String id) {
+    String system = "terminology.hl7.org/CodeSystem/medicationrequest-category";
```
jblefkove-va, line 201: Should this be a constant?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestControllerTest.java
```diff
@@ -74,7 +76,80 @@ R4MedicationRequestController controller() {
...
+  }
+
+  @Test
+  public void correctSuffixDetermination() {
+    var fhir = MedicationRequestSamples.R4.create();
+    String system = "terminology.hl7.org/CodeSystem/medicationrequest-category";
+    MedicationRequest req = fhir.medicationRequestFromMedicationOrder();
+    assertThat(req).isNotNull();
+
+    req = fhir.medicationRequestFromMedicationOrder("123:");
```
jblefkove-va, line 91: there should also be a test case for an unknown suffix, e.g. `123:XXX`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestController.java
```diff
@@ -61,16 +63,24 @@
 
   private final WitnessProtection witnessProtection;
 
+  @NonNull private final Pattern outPattern;
+
+  @NonNull private final Pattern inPattern;
```
jhulbert-va, line 68: I don't think these need to be nonnull in the controller, just the transformer.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestControllerTest.java
```diff
@@ -74,7 +76,83 @@ R4MedicationRequestController controller() {
...
+    assertThat(req.category()).isEqualTo(null);
+
+    req = fhir.medicationRequestFromMedicationOrder("123:XXX");
+    assertThat(req.category()).isEqualTo(null);
+
+    req = fhir.medicationRequestFromMedicationOrder("123:I");
+    assertThat(req.category())
+        .isEqualTo(
+            List.of(
```
jhulbert-va, line 100: Why not pull this out as `var inpatient` so you don't have to build it twice?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestControllerTest.java
```diff
@@ -74,7 +76,83 @@ R4MedicationRequestController controller() {
...
         medicationOrderRepository,
         medicationStatementRepository,
-        WitnessProtection.builder().identityService(ids).build());
+        WitnessProtection.builder().identityService(ids).build(),
+        ".*:(O|FP)",
+        ".*:(I|FPI)");
+  }
+
+  @Test
+  public void correctSuffixDetermination() {
```
jhulbert-va, line 85: Pretty sure all this test does is test your samples, not your controller/transformer code.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestController.java
jblefkove-va: The `@NonNull` annotation is less useful on controllers, because they are Spring components; the container manages their creation, and Spring is unlikely to miss setting a field. The transformers are just regular Java classes though; we control their creation, so the `@NonNull` ensures an obvious runtime error if you forget to set a required field.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestControllerTest.java
```diff
@@ -74,7 +76,89 @@ R4MedicationRequestController controller() {
...
+    var fhir = MedicationRequestSamples.R4.create();
+    String system = "https://www.hl7.org/fhir/codesystem-medicationrequest-category.html";
+    MedicationRequest req = fhir.medicationRequestFromMedicationOrder();
+    assertThat(req).isNotNull();
+
+    req = fhir.medicationRequestFromMedicationOrder("123:");
+    assertThat(req.category()).isEqualTo(null);
```
jhulbert-va, line 92: This and a few others are still testing your samples and are unnecessary.

jblefkove-va: Good catch Josh... the only asserts that are testing production code are the ones that look like this:
```
assertThat(controller().transformMedicationOrderToMedicationRequest(...).category()).isEqualTo(...);
```

It would also be clearer to move these tests to `R4MedicationRequestFromMedicationOrderTransformerTest` to test `R4MedicationRequestFromMedicationOrderTransformer.parseCategory()` directly

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestControllerTest.java
```diff
@@ -74,7 +76,83 @@ R4MedicationRequestController controller() {
...
+
+  @Test
+  public void correctSuffixDetermination() {
+    var fhir = MedicationRequestSamples.R4.create();
+    String system = "https://www.hl7.org/fhir/codesystem-medicationrequest-category.html";
+
+    assertThat(
+            controller()
+                .transformMedicationOrderToMedicationRequest(
+                    MedicationOrderSamples.Datamart.create().medicationOrder("123:", "1234"))
```
jhulbert-va, line 92: One more test. What about MedOrd id `1234` with no `:`?

jhulbert-va (approved):

PR #608, opened 2020-10-14
Removed ID service from Data Query Tests

that-guy-is-it:


jblefkove-va:
Update the tests pom so that Id service does not get launched as part of the integration tests

jblefkove-va:
I suspect the dependencies on ids-api and the ids app in the pom are also safe to remove.

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinition.java
```diff
@@ -8,7 +8,7 @@
 @Value
 @Builder
 public final class SystemDefinition {
-  @NonNull ServiceDefinition ids;
+  //  @NonNull ServiceDefinition ids;
```
jblefkove-va, line 11: Remove this?

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/TestClients.java
```diff
@@ -26,7 +26,6 @@ public static TestClient dstu2DataQuery() {
   /** ID service test client. */
   public static TestClient ids() {
     return BasicTestClient.builder()
-        .service(SystemDefinitions.systemDefinition().ids())
```
jblefkove-va, line 29: This whole method should be removed

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinition.java
bschofield-va: this is how cruft builds up

bschofield-va (commented):
`IdsRegistrar.register(...)` also needs to be deleted

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/IdRegistrar.java
```diff
@@ -86,29 +62,6 @@ private TestIds registerCdwIds() {
...
-            .medicationStatement(findUuid(registrations, medicationStatement))
-            .observation(findUuid(registrations, observation))
-            .organization(findUuid(registrations, organizataion))
-            .patient(findUuid(registrations, patient))
-            .practitioner(findUuid(registrations, practitioner))
-            .procedure(findUuid(registrations, procedure))
-            .build();
-    log.info("Using {}", publicIds);
-    return publicIds;
+    return null;
```
jblefkove-va, line 65: What else from this method can be removed?

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/IdRegistrar.java
```diff
@@ -24,31 +21,10 @@
...
-        TestClients.ids()
-            .post("/api/v1/ids", singletonList(identity))
-            .expect(201)
-            .expectListOf(Registration.class);
-    return findUuid(registrations, identity);
-  }
-
   private TestIds registerCdwIds() {
     TestIds cdwIds = system().cdwIds();
     if (cdwIds.publicIds()) {
```
jblefkove-va, line 30: This `if` statement seems to be the only part of this method that actually runs. Can we determine why?

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/IdRegistrar.java
```diff
@@ -24,91 +18,9 @@
...
-    List<Registration> registrations =
-        TestClients.ids()
-            .post("/api/v1/ids", singletonList(identity))
-            .expect(201)
-            .expectListOf(Registration.class);
-    return findUuid(registrations, identity);
-  }
-
   private TestIds registerCdwIds() {
     TestIds cdwIds = system().cdwIds();
```
jblefkove-va, line 22: This is the only useful line left in this method. (Or in this whole class.) Let's inline this in `ResourceVerifier` and remove `IdRegistrar` altogether.

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinition.java
```diff
@@ -8,8 +8,6 @@
 @Value
 @Builder
 public final class SystemDefinition {
-  @NonNull ServiceDefinition ids;
```
jblefkove-va, line 11: The removal of the always-true `publicIds` field from `TestIds` points out that `cdwIds` is misnamed here. CDW IDs are private; if the test IDs can be public, then `cdewIds` is not an appropriate name; since we've now determined that they are *always* public, let's call the field `publicIds`.

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -208,7 +203,6 @@ private static SystemDefinition prod() {
   private static TestIds productionCdwIds() {
```
jblefkove-va, line 203: Rename this to `productionIds`.

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -188,27 +183,29 @@ private static Observations observations() {
...
         .stu3DataQuery(
             serviceDefinition("stu3", url, 443, magicAccessToken(), "/services/fhir/v0/stu3/"))
         .r4DataQuery(serviceDefinition("r4", url, 443, magicAccessToken(), "/services/fhir/v0/r4/"))
         .internalDataQuery(serviceDefinition("internal", url, 443, null, "/not-available/"))
-        .cdwIds(productionCdwIds())
+        .publicIds(productionIds())
         .build();
   }
 
+  private static Procedures productionCdwProcedures() {
```
jblefkove-va, line 196: I guess this should be renamed as well

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -272,7 +265,6 @@ private static Procedures productionCdwProcedures() {
         .family("ACOSTA")
         .given("SAID R")
         .npi("http://hl7.org/fhir/sid/us-npi|1013904481")
-        // Valid ID will need to be added when test is available
```
jblefkove-va, line 275: formatter ate this comment

jblefkove-va (commented):
:shipit:

jblefkove-va (approved):

PR #612, opened 2020-10-30
API 3079 vulcanize prototype

bschofield-va:
Vulcanized Diagnostic Report with testing examples.


data-query/src/main/java/gov/va/api/health/dataquery/service/config/FhirMediaTypesConfig.java
```diff
@@ -27,8 +27,8 @@
   private final int defaultCount;
 
   public FhirMediaTypesConfig(
-      @Value("${count-parameter.max-count:20}") int maxCount,
-      @Value("${count-parameter.default-count:15}") int defaultCount) {
+      @Value("${data-query.max-page-size:20}") int maxCount,
+      @Value("${data-query.default-page-size:15}") int defaultCount) {
```
jhulbert-va, line 31: This will require a DU update. Setting the page size to 30 is happening there.

bschofield-va: The DU update is already in place

eclendenning-va (approved):

PR #613, opened 2020-11-03
initial vulcanized practitionerrole controller

apeirats-va:
This is my first pass at implementing the R4 PractitionerRole controller through Vulcan.

### DIFF
--- a/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
+++ b/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
@@ -28,14 +28,4 @@ public class PractitionerRoleIT {
         test(400, OperationOutcome.class, "PractitionerRole/"),
         test(400, OperationOutcome.class, "PractitionerRole?blah=123"));
   }
-
-  @Test
-  public void notImplementedParameters() {
-    verifier.verifyAll(
-        // Throws NotImplemented exception
-        test(501, OperationOutcome.class, "PractitionerRole?identifier=123"),
-        test(501, OperationOutcome.class, "PractitionerRole?specialty=system|code"),
-        test(501, OperationOutcome.class, "PractitionerRole?practitioner.identifier=system|code"),
-        test(501, OperationOutcome.class, "PractitionerRole?practitioner.name=Doe"));
-  }
 }
--- a/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
+++ b/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
@@ -33,6 +33,7 @@ public class PractitionerRoleIT {
   public void notImplementedParameters() {
     verifier.verifyAll(
         // Throws NotImplemented exception
+        test(501, OperationOutcome.class, "PractitionerRole?identifier=123"),
         test(501, OperationOutcome.class, "PractitionerRole?specialty=system|code"),
         test(501, OperationOutcome.class, "PractitionerRole?practitioner.identifier=system|code"),
         test(501, OperationOutcome.class, "PractitionerRole?practitioner.name=Doe"));
--- a/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
+++ b/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
@@ -14,42 +14,27 @@ public class PractitionerRoleIT {
     verifier.verifyAll(
         test(200, PractitionerRole.class, "PractitionerRole/{id}", verifier.ids().practitioner()),
         test(404, OperationOutcome.class, "PractitionerRole/{id}", verifier.ids().unknown()),
-        // search by identifier
+        // search by _id
         test(
             200,
             PractitionerRole.Bundle.class,
-            "PractitionerRole?identifier={npi}",
-            verifier.ids().practitioners().npi()),
-        // search by practitioner.identifier
-        test(
-            200,
-            PractitionerRole.Bundle.class,
-            "PractitionerRole?practitioner.identifier={npi}",
-            verifier.ids().practitioners().npi()),
-        // search by practitioner.name Family name
-        test(
-            200,
-            PractitionerRole.Bundle.class,
-            "PractitionerRole?practitioner.name={name}",
-            verifier.ids().practitioners().family()));
+            "PractitionerRole?_id={id}",
+            verifier.ids().practitioner()));
   }
 
   @Test
   public void malformed() {
     verifier.verifyAll(
         test(400, OperationOutcome.class, "PractitionerRole/"),
-        test(400, OperationOutcome.class, "PractitionerRole?blah=123"),
-        // parameters not specified together
-        test(
-            400,
-            OperationOutcome.class,
-            "PractitionerRole?practitioner.name=Somename&practitioner.identifier=123"));
+        test(400, OperationOutcome.class, "PractitionerRole?blah=123"));
   }
 
   @Test
-  public void searchBySpecialty() {
+  public void notImplementedParameters() {
     verifier.verifyAll(
         // Throws NotImplemented exception
-        test(501, OperationOutcome.class, "PractitionerRole?specialty=system|code"));
+        test(501, OperationOutcome.class, "PractitionerRole?specialty=system|code"),
+        test(501, OperationOutcome.class, "PractitionerRole?practitioner.identifier=system|code"),
+        test(501, OperationOutcome.class, "PractitionerRole?practitioner.name=Doe"));
   }
 }
--- a/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
+++ b/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
@@ -1,8 +1,11 @@
 package gov.va.api.health.dataquery.tests.r4;
 
+import static gov.va.api.health.sentinel.EnvironmentAssumptions.assumeEnvironmentIn;
+
 import gov.va.api.health.dataquery.tests.ResourceVerifier;
 import gov.va.api.health.r4.api.resources.OperationOutcome;
 import gov.va.api.health.r4.api.resources.PractitionerRole;
+import gov.va.api.health.sentinel.Environment;
 import lombok.experimental.Delegate;
 import org.junit.jupiter.api.Test;
 
@@ -24,6 +27,7 @@ public class PractitionerRoleIT {
 
   @Test
   public void malformed() {
+    assumeEnvironmentIn(Environment.LOCAL);
     verifier.verifyAll(
         test(400, OperationOutcome.class, "PractitionerRole/"),
         test(400, OperationOutcome.class, "PractitionerRole?blah=123"));
--- a/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
+++ b/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
@@ -1,0 +1,55 @@
+package gov.va.api.health.dataquery.tests.r4;
+
+import gov.va.api.health.dataquery.tests.ResourceVerifier;
+import gov.va.api.health.r4.api.resources.OperationOutcome;
+import gov.va.api.health.r4.api.resources.PractitionerRole;
+import lombok.experimental.Delegate;
+import org.junit.jupiter.api.Test;
+
+public class PractitionerRoleIT {
+  @Delegate ResourceVerifier verifier = ResourceVerifier.r4();
+
+  @Test
+  public void basic() {
+    verifier.verifyAll(
+        test(200, PractitionerRole.class, "PractitionerRole/{id}", verifier.ids().practitioner()),
+        test(404, OperationOutcome.class, "PractitionerRole/{id}", verifier.ids().unknown()),
+        // search by identifier
+        test(
+            200,
+            PractitionerRole.Bundle.class,
+            "PractitionerRole?identifier={npi}",
+            verifier.ids().practitioners().npi()),
+        // search by practitioner.identifier
+        test(
+            200,
+            PractitionerRole.Bundle.class,
+            "PractitionerRole?practitioner.identifier={npi}",
+            verifier.ids().practitioners().npi()),
+        // search by practitioner.name Family name
+        test(
+            200,
+            PractitionerRole.Bundle.class,
+            "PractitionerRole?practitioner.name={name}",
+            verifier.ids().practitioners().family()));
+  }
+
+  @Test
+  public void malformed() {
+    verifier.verifyAll(
+        test(400, OperationOutcome.class, "PractitionerRole/"),
+        test(400, OperationOutcome.class, "PractitionerRole?blah=123"),
+        // parameters not specified together
+        test(
+            400,
+            OperationOutcome.class,
+            "PractitionerRole?practitioner.name=Somename&practitioner.identifier=123"));
+  }
+
+  @Test
+  public void searchBySpecialty() {
+    verifier.verifyAll(
+        // Throws NotImplemented exception
+        test(501, OperationOutcome.class, "PractitionerRole?specialty=system|code"));
+  }
+}
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DataQueryService.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DataQueryService.java
@@ -11,7 +11,6 @@ import gov.va.api.health.r4.api.MetadataApi;
 import gov.va.api.health.r4.api.ObservationApi;
 import gov.va.api.health.r4.api.OrganizationApi;
 import gov.va.api.health.r4.api.PatientApi;
-import gov.va.api.health.r4.api.PractitionerRoleApi;
 import gov.va.api.health.r4.api.ProcedureApi;
 import io.swagger.v3.oas.annotations.ExternalDocumentation;
 import io.swagger.v3.oas.annotations.OpenAPIDefinition;
@@ -115,5 +114,4 @@ public interface R4DataQueryService
         ObservationApi,
         OrganizationApi,
         PatientApi,
-        PractitionerRoleApi,
         ProcedureApi {}
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DataQueryService.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DataQueryService.java
@@ -11,6 +11,7 @@ import gov.va.api.health.r4.api.MetadataApi;
 import gov.va.api.health.r4.api.ObservationApi;
 import gov.va.api.health.r4.api.OrganizationApi;
 import gov.va.api.health.r4.api.PatientApi;
+import gov.va.api.health.r4.api.PractitionerRoleApi;
 import gov.va.api.health.r4.api.ProcedureApi;
 import io.swagger.v3.oas.annotations.ExternalDocumentation;
 import io.swagger.v3.oas.annotations.OpenAPIDefinition;
@@ -114,4 +115,5 @@ public interface R4DataQueryService
         ObservationApi,
         OrganizationApi,
         PatientApi,
+        PractitionerRoleApi,
         ProcedureApi {}
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
@@ -55,7 +55,7 @@ public class R4PractitionerRoleController {
     return (r) -> {
       if (r.request().getParameter(parameter) != null) {
         throw new ResourceExceptions.NotImplemented(
-            parameter + " search param is not yet implemented");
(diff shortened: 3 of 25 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,174 @@
...
+  private PractitionerRepository repository;
+
+  /** Vulcan rule for unimplemented parameter. */
+  public static Rule parameterNotImplemented(String parameter) {
+    return (r) -> {
+      if (r.request().getParameter(parameter) != null) {
+        throw new ResourceExceptions.NotImplemented("not-implemented");
+      }
+    };
+  }
```
apeirats-va, line 64: Can probably move method the method above to Vulcan's Rules class. @bschofield-va any opinions?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,174 @@
...
+
+  private VulcanConfiguration<PractitionerEntity> configuration() {
+    return VulcanConfiguration.forEntity(PractitionerEntity.class)
+        .paging(
+            linkProperties.pagingConfiguration(
+                "PractitionerRole", PractitionerEntity.naturalOrder()))
+        .mappings(
+            Mappings.forEntity(PractitionerEntity.class)
+                .value("_id", "cdwId", witnessProtection::toCdwId)
```
apeirats-va, line 73: Need to add a mapping for `practitioner.name`

jhulbert-va (commented):
Are the swagger/metadata/IT changes going to be in a separate PR?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
bschofield-va: Do we anticipate needing it more than once? If so, yes.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,174 @@
...
+        .bundling(
+            Bundling.newBundle(PractitionerRole.Bundle::new)
+                .newEntry(PractitionerRole.Entry::new)
+                .linkProperties(linkProperties)
+                .build())
+        .build();
+  }
+
+  /** Validate identifier token. */
+  public boolean tokenIdentifierIsSupported(TokenParameter token) {
```
jhulbert-va, line 130: Can we rename these methods? I think `tokenIdentifier*` is a little bit misleading. Instead, why not something like `practitionerIdentifierTokenIsSupported`? I think that creates better separation between the token itself and the parameter it represents.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,174 @@
...
+                .token(
+                    "practitioner.identifier",
+                    "npi",
+                    this::tokenIdentifierIsSupported,
+                    this::tokenIdentifierValue)
+                .get())
+        .defaultQuery(returnNothing())
+        .rule(
+            atLeastOneParameterOf(
+                "specialty", "practitioner.identifier", "practitioner.name", "_id"))
```
jhulbert-va, line 83: We should also be supporting the plain old `identifier` param.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,174 @@
...
+
+  private final LinkProperties linkProperties;
+
+  private PractitionerRepository repository;
+
+  /** Vulcan rule for unimplemented parameter. */
+  public static Rule parameterNotImplemented(String parameter) {
+    return (r) -> {
+      if (r.request().getParameter(parameter) != null) {
+        throw new ResourceExceptions.NotImplemented("not-implemented");
```
jhulbert-va, line 61: This should probably have more information on _what_ parameter isn't implemented.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
bschofield-va: I don't know how to interpret this. Does this mean you are still making changes in this PR or this will be implemented in a different PR?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
bschofield-va: because it nicely sorts the tokenXXX methods together. I prefer this style because of the natural grouping of methods.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleControllerTest.java
```diff
@@ -0,0 +1,179 @@
...
+                "http://fonzy.com/r4/PractitionerRole?practitioner.identifier=pr1",
+                9,
+                15));
+    var applied = bundler.apply(vr);
+    assertThat(applied).isEqualTo(expected);
+  }
+
+  @ParameterizedTest
+  @CsvSource({
+    "12345,false",
```
jhulbert-va, line 149: Shouldn't this token format be supported?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
jhulbert-va: Fair enough. I think this should be `tokenPractitionerIdentifier*` still. The plain old `identifier` parameter is also valid, but not a token.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
apeirats-va: I'm making/testing the change now and will update the PR.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
apeirats-va: The grep for it suggests it's not used anywhere other than PractitionerRole, but I think there may be a case for it in the future.
However I'm happy to just leave it here since, as it stands, it seems to be just a one-of.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
apeirats-va: I just added support for a plain `identifier` value that represents just the NPI, and renamed this method to make it clear it represents the `practitioner.identifier` token.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleControllerTest.java
apeirats-va: Followed logic in STU3 implementation, which seems to bail if no system is specified or if the system does not match the exact `us-npi` system.

For the case of R4, if no system is specified should we just assume NPI?

I did add a `identifier` which just takes a NPI value, so that's also an option for the consumer.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
jhulbert-va: In the future, GitHub has a draft PR feature if you want to put out a PR while you're still working on features.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleControllerTest.java
jhulbert-va: @jblefkove-va was there a reason for that in the stu3 implementation?

jblefkove-va: I suspect the STU3 implementation is wrong; it predates our current understanding of token searches.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
apeirats-va: Nvm, moved to API-3444

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleControllerTest.java
apeirats-va: Yanked out NPI searches and moved the identifier/name search implementations to API-3444.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,180 @@
...
+  private VulcanConfiguration<PractitionerEntity> configuration() {
+    return VulcanConfiguration.forEntity(PractitionerEntity.class)
+        .paging(
+            linkProperties.pagingConfiguration(
+                "PractitionerRole", PractitionerEntity.naturalOrder()))
+        .mappings(
+            Mappings.forEntity(PractitionerEntity.class)
+                .value("_id", "cdwId", witnessProtection::toCdwId)
+                .value("identifier", "npi")
+                .string("practitioner.name", "familyName")
```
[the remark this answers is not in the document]
apeirats-va: Moved name search implementation to API-3444

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,144 @@
...
+ * https://build.fhir.org/ig/HL7/US-Core-R4/StructureDefinition-us-core-practitionerrole.html for
+ * implementation details.
+ */
+@Validated
+@RestController
+@SuppressWarnings("WeakerAccess")
+@RequestMapping(
+    value = "/r4/PractitionerRole",
+    produces = {"application/json", "application/fhir+json"})
+@AllArgsConstructor(onConstructor = @__({@Autowired}))
```
jhulbert-va, line 45: Can you make this `onConstructor_ = @Autowired`? I think it reads better that way.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
```diff
@@ -0,0 +1,40 @@
...
+        // search by _id
+        test(
+            200,
+            PractitionerRole.Bundle.class,
+            "PractitionerRole?_id={id}",
+            verifier.ids().practitioner()));
+  }
+
+  @Test
+  public void malformed() {
```
jhulbert-va, line 26: We don't need the IT's to test the malformed requests. Just each path for reads/searches.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,144 @@
...
+    };
+  }
+
+  private VulcanConfiguration<PractitionerEntity> configuration() {
+    return VulcanConfiguration.forEntity(PractitionerEntity.class)
+        .paging(
+            linkProperties.pagingConfiguration(
+                "PractitionerRole", PractitionerEntity.naturalOrder()))
+        .mappings(
+            Mappings.forEntity(PractitionerEntity.class)
```
jhulbert-va, line 69: What about plain old `identifier`?

jhulbert-va (commented):
@bschofield-va @ramosan-lit In regards to the swagger and metadata, I'm not sure whether we want to make that public until the conversation about scopes and scope types is resolved. Thoughts?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
apeirats-va: That's still a token in this case, right? If so, then API-3444

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
apeirats-va: I think I noticed some tests check them, some don't. So I just threw them in but in their own method.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
apeirats-va: I could. I just copied it from another controller. I had not realized the syntax you're suggesting was even an option. I tried it out locally and it seems to work fine, so I'll get that committed.

jhulbert-va: Lombok originally only supported the first way. But the new way is a lot more readable.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
jhulbert-va: If you want to do that, add the local only assumption. Malformed request tests are unnecessary during a deployment.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
jhulbert-va: Cool. That being said, I'll resolve this conversation.

data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DataQueryService.java
```diff
@@ -114,4 +115,5 @@
         ObservationApi,
         OrganizationApi,
         PatientApi,
+        PractitionerRoleApi,
```
jhulbert-va, line 118: If all the other search params won't be made available until API-3444. I would prefer we don't update the metadata and swagger right now.

apeirats-va: Ok I'll revert that commit.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,144 @@
...
+                .value("_id", "cdwId", witnessProtection::toCdwId)
+                .get())
+        .defaultQuery(returnNothing())
+        .rule(
+            atLeastOneParameterOf(
+                "_id", "practitioner.name", "practitioner.identifier", "specialty"))
+        .rule(parameterNotImplemented("practitioner.name"))
```
jhulbert-va, line 76: Last thing, lets add plain `identifier` to these rules so if its requested there's a not implemented.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,144 @@
...
+  private final LinkProperties linkProperties;
+
+  private PractitionerRepository repository;
+
+  /** Vulcan rule for unimplemented parameter. */
+  public static Rule parameterNotImplemented(String parameter) {
+    return (r) -> {
+      if (r.request().getParameter(parameter) != null) {
+        throw new ResourceExceptions.NotImplemented(
+            parameter + " search param is not yet implemented");
```
bschofield-va, line 58: Just a nit, but I'd prefer to avoid casual abbreviations in error messages seen by end users.
_search parameter_ vs _search param_

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
bschofield-va: Just a note... all of these rules could be combined as follows and virtually achieve the same result - Disallow these parameters.
```
.rule(atLeastOneParameterOf("_id"))
.rule(forbidUnknownParameter())
```

I don't really understand what returning 503 - Not Implemented is buying us over just returning 400 - Bad Request. We don't specifically 503 all 13 defined query parameters for PractitionerRole. Why just these 3?

apeirats-va: 501* is Not Implemented.

It's technically a valid request from the user's perspective so returning 400 feels misleading.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
apeirats-va: This message only makes its way to our logs and not the response. I'll reword it to `parameter` anyways.

jhulbert-va:
Don't forget the `IcnIncludesMajig`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
bschofield-va: but 400 for the rest these searches is ok? http://hl7.org/fhir/R4/practitionerrole.html#search
I value _consistency_ but will defer to Ramos

apeirats-va: Hm, I had not even seen that list of parameters TBH.
I based it only on the mandatory ones called out here: http://hl7.org/fhir/us/core/StructureDefinition-us-core-practitionerrole.html#mandatory-search-parameters

I agree with the consistency comment. For the sake of this PR and just making the resource reads available, I don't want to keep flip-flopping on search parameters, but I'm happy to let you make the final call.

API-3444 should finalize the implementation of the required parameters.

apeirats-va: Following conversation on slack, the unsupported search params are now ignored and returning 400.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -0,0 +1,127 @@
...
+        .paging(
+            linkProperties.pagingConfiguration(
+                "PractitionerRole", PractitionerEntity.naturalOrder()))
+        .mappings(
+            Mappings.forEntity(PractitionerEntity.class)
+                .value("_id", "cdwId", witnessProtection::toCdwId)
+                .get())
+        .defaultQuery(returnNothing())
+        .rule(atLeastOneParameterOf("_id"))
+        .rule(forbidUnknownParameters())
```
bschofield-va, line 62: ^ remove this rule

apeirats-va: Since FHIR says unknown parameters shoudl be simply ignored, should the rule `forbidUnknownParameters` not be used at all by R4 resource controllers?

We had talked about DQ rejecting unknown parameters, but I'm guessing that's no longer the case.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleControllerTest.java
```diff
@@ -0,0 +1,143 @@
...
+import lombok.SneakyThrows;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class R4PractitionerRoleControllerTest {
```
jhulbert-va, line 37: ValidRequests test

jhulbert-va (approved):

PR #614, opened 2020-11-05
API-3430-datamart-device

jhulbert-va:
# [API-3430](https://vajira.max.gov/browse/API-3430)

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/device/DatamartDevice.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/device/DatamartDevice.java
@@ -3,6 +3,7 @@ package gov.va.api.health.dataquery.service.controller.device;
 import gov.va.api.health.dataquery.service.controller.datamart.DatamartCoding;
 import gov.va.api.health.dataquery.service.controller.datamart.DatamartReference;
 import gov.va.api.health.dataquery.service.controller.datamart.HasReplaceableId;
+import java.util.Optional;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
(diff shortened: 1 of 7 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/device/DatamartDevice.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class DatamartDevice implements HasReplaceableId {
```
bschofield-va, line 16: Are we guaranteed all of these fields?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/device/DeviceRepository.java
```diff
@@ -0,0 +1,14 @@
...
+import gov.va.api.health.autoconfig.logging.Loggable;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.repository.PagingAndSortingRepository;
+import org.springframework.transaction.annotation.Isolation;
+import org.springframework.transaction.annotation.Transactional;
+
+@Loggable
+@Transactional(isolation = Isolation.READ_UNCOMMITTED)
+public interface DeviceRepository extends PagingAndSortingRepository<DeviceEntity, String> {
```
bschofield-va, line 12: You'll also want to extend `JpaSpecificationExecutor` for Vulcan

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/device/DatamartDevice.java
jhulbert-va: > The JSON function would allow all NULL values to be returned from the function. I think from our ETL process, we should always have a valid Patient, so the Patient fields should not be NULL. And if the Device table is correctly populated, the last 6 fields should not be NULL.

eclendenning-va (approved):
LGTM

PR #618, opened 2020-11-09
API-3429-r4-device-controller

jhulbert-va:
# [API-3429](https://vajira.max.gov/browse/API-3429)


jhulbert-va:
Hold off on merging. The production ID for the IT's will probably need to be updated.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/device/R4DeviceController.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private final LinkProperties linkProperties;
+
+  private VulcanConfiguration<DeviceEntity> configuration() {
+    return VulcanConfiguration.forEntity(DeviceEntity.class)
+        .paging(linkProperties.pagingConfiguration("Device", DeviceEntity.naturalOrder()))
+        .mappings(
+            Mappings.forEntity(DeviceEntity.class)
+                .value("_id", "cdwId", witnessProtection::toCdwId)
+                // ToDo -- cdwId system for a token search?
```
bschofield-va, line 53: This is universally true for all resources... let's make an item in the backlog to decide how we want (if at all) to support a system for `_id` searches and remove the comment

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/DeviceIT.java
```diff
@@ -0,0 +1,40 @@
...
+        test(404, OperationOutcome.class, "Device/{id}", verifier.ids().unknown()),
+        test(200, Device.Bundle.class, "Device?patient={patientIcn}", verifier.ids().patient()));
+  }
+
+  @Test
+  void searchByIdentifier() {
+    assumeEnvironmentIn(Environment.LOCAL);
+    verifier.verifyAll(
+        test(200, Device.Bundle.class, "Device?_id={id}", verifier.ids().device()),
+        test(200, Device.Bundle.class, "Device?_id={id}", verifier.ids().unknown()),
```
bschofield-va, line 29: How hard would this be to to have an empty verification, say by adding a version of the `test` method accepts a predicate, e.g.
```
test(200, b->b.entry().isEmpty(), Device.Bundle.class, "Device?_id={id}", verifier.ids().unknown())
```

bschofield-va (dismissed):

bschofield-va (approved):

PR #619, opened 2020-11-12
API-2972 Update Practitioner IT to Local

that-guy-is-it:


data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerIT.java
```diff
@@ -14,6 +14,7 @@
 
   @Test
   public void basic() {
+    assumeEnvironmentIn(Environment.LOCAL);
```
jhulbert-va, line 17: Take a look at `DeviceIT` and fix this accordingly. Adding this line, effectively turns off all IT's for R4 Practitioner. We still want some of the tests here to run outside of local.

jhulbert-va (changes requested):

jhulbert-va (approved):

PR #621, opened 2020-11-17
Vulcanized read raw now supports 'x-va-includes-icn: NONE' when records do not apply to a patient

bschofield-va:


jhulbert-va (approved):
LGTM

PR #623, opened 2020-11-23
API-3710: swagger and capability/conformance verification

jjabando1:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/Dstu2MetadataController.java
```diff
@@ -112,64 +99,35 @@
...
-      default:
-        throw noSearchParamsForConformanceStatementTypeException();
-    }
-  }
-
   private Set<SearchParam> patientSearchParams() {
     switch (properties.getStatementType()) {
       case PATIENT:
-        return singleton(SearchParam.ID);
+        return ImmutableSet.of(
```
ilaflamme-va, line 111: We do not want to advertise our patient searching. Only search by ID.

ilaflamme-va (changes requested):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/Dstu2MetadataController.java
```diff
@@ -194,40 +147,81 @@ Conformance read() {
...
+                    ImmutableSet.of(
+                        SearchParam.CATEGORY,
+                        SearchParam.CLINICAL_STATUS,
+                        SearchParam.ID,
+                        SearchParam.IDENTIFIER,
+                        SearchParam.PATIENT))
                 .build(),
             support("DiagnosticReport")
                 .documentation(DIAGNOSTICREPORT_HTML)
-                .search(diagnosticReportSearchParams())
```
jhulbert-va, line 205: Instead of removing the method entirely, I think that the switch should just fall through.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/Dstu2MetadataController.java
```diff
@@ -90,10 +89,13 @@
   private Set<SearchParam> conditionSearchParams() {
     switch (properties.getStatementType()) {
       case PATIENT:
```
bschofield-va, line 91: always document fall through so everyone knows it was intentional. Your IDE should have been warning you.
```
case Patient:
  // fall through
case CLINICIAN:
```

bschofield-va: Actually... what's the point of the case statement at all.

jhulbert-va: That's on me. I asked him to keep the case statements for any resource with more than id/identifier. I feel it's more clear as to the intent.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/Dstu2MetadataController.java
```diff
@@ -194,7 +211,8 @@ Conformance read() {
     return Stream.of(
             support("AllergyIntolerance")
                 .documentation(ALLERGYINTOLERANCE_HTML)
-                .search(ImmutableSet.of(SearchParam.PATIENT))
+                .search(
```
bschofield-va, line 214: This bothers me now... previously we used methods (e.g., `conditionSearchParams()`) to decide what parameters apply to what mode. But now those methods do no logic. They _always_ return the same set, and _now_ we have a two different ways of doing the same thing. For resources that have constant parameters regardless of the mode (patient/clinician) we sometimes build the set in place, sometime call a method to build a set.

Pick one way and do it. Either inline the method return values here or move all set creation to methods. If you keep the methods, you need to remove the case statements from all them that do not make any decisions.

bschofield-va (dismissed):
do not like methods for search parameters with unnecessary switch statements

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/Dstu2MetadataController.java
```diff
@@ -306,6 +310,7 @@ private Software software() {
     GENDER("gender", SearchParamType.token),
     GIVEN("given", SearchParamType.string),
     ID("_id", SearchParamType.string),
+    IDENTIFIER("identifier", SearchParamType.string),
```
jhulbert-va, line 313: Identifier is _technically_ a token.

eclendenning-va: while this is true, isnt the point of the story to bring it into line with what we currently support? We currently don't support that as a token.

jhulbert-va: Gender is marked as a token, but we only support it as a string as well. ¯\_(ツ)_/¯

eclendenning-va: Thats a good point. I think `gender` should also be a string in the swagger as well then.

jhulbert-va: I don't know that I agree. If we were to do that, then the metadata basically says "we're not fhir compliant".

eclendenning-va: I think its a little bit of false advertising otherwise. I'll defer to @ramosan-lit .

eclendenning-va: After re-reading the definition of the string search without `:exact` I think Josh is correct.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/R4MetadataController.java
```diff
@@ -258,6 +271,7 @@ private Software software() {
     GENDER("gender", SearchParamType.token),
     GIVEN("given", SearchParamType.string),
     ID("_id", SearchParamType.string),
```
eclendenning-va, line 273: `category` and `_id` should also be tokens

jhulbert-va: I thought `_id` was a string and `identifier` is a token?

https://www.hl7.org/fhir/search.html#id

jhulbert-va: https://www.hl7.org/fhir/resource.html#search

jhulbert-va: @jjabando1 what about dstu2 category?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/Dstu2MetadataController.java
```diff
@@ -298,14 +302,15 @@ private Software software() {
...
     BIRTH_DATE("birthdate", SearchParamType.date),
-    CATEGORY("category", SearchParamType.string),
+    CATEGORY("category", SearchParamType.token),
     CLINICAL_STATUS("clinicalstatus", SearchParamType.token),
     CODE("code", SearchParamType.token),
     DATE("date", SearchParamType.date),
     FAMILY("family", SearchParamType.string),
     GENDER("gender", SearchParamType.token),
     GIVEN("given", SearchParamType.string),
     ID("_id", SearchParamType.string),
```
jhulbert-va, line 312: `_id` is a token.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/R4MetadataController.java
```diff
@@ -250,14 +263,15 @@ private Software software() {
...
     CLINICAL_STATUS("clinical-status", SearchParamType.token),
     CODE("code", SearchParamType.token),
     DATE("date", SearchParamType.date),
     FAMILY("family", SearchParamType.string),
     GENDER("gender", SearchParamType.token),
     GIVEN("given", SearchParamType.string),
-    ID("_id", SearchParamType.string),
+    ID("_id", SearchParamType.token),
+    IDENTIFIER("identifier", SearchParamType.token),
     INTENT("intent", SearchParamType.string),
```
jhulbert-va, line 275: `intent` is a token.

https://www.hl7.org/fhir/medicationrequest.html#search

jhulbert-va (approved):
LGTM

PR #645, opened 2021-01-14
API-4379: datamartOrganization facilityId

jjabando1:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/organization/DatamartOrganization.java
```diff
@@ -45,7 +45,7 @@
 
   private Optional<DatamartReference> partOf;
 
-  private Optional<Facility> facility;
+  private Optional<Facility> facilityId;
```
jhulbert-va, line 48: The class should be `FacilitiyID` as well.

jhulbert-va (approved):
LGTM

PR #649, opened 2021-01-28
new appointment entity and accompanying datamart object structure

ilaflamme-va:
THIS IS PART 2 OF: https://vajira.max.gov/browse/API-2563

see [here](https://github.com/department-of-veterans-affairs/health-apis-datamart-synthetic-records/pull/90) for the DB changes in PART 1


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/DatamartAppointmentTest.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.SneakyThrows;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class DatamartAppointmentTest {
+  private final DatamartAppointment expected = AppointmentSamples.Datamart.create().appointment();
+
+  @SneakyThrows
+  @ParameterizedTest
+  @ValueSource(strings = {"datamart-appointment-v1.json"})
```
jhulbert-va, line 15: There should also be a file called `datamart-appointment.json` that will always have the latest version (for the initial object, they'll be the same).

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/DatamartAppointment.java
```diff
@@ -0,0 +1,59 @@
...
+  private Integer minutesDuration;
+
+  private String created;
+
+  private Optional<String> comment;
+
+  private List<DatamartReference> participant;
+
+  public Optional<String> appointmentType() {
+    return lazyGetter(appointmentType);
```
jhulbert-va, line 46: I think this needs to be:
```
appointmentType = lazyGetter(appointmentType);
return appointmentType;
```

Lazy getters usually set the fields before returning them:
```
  public Optional<String> code() {
    if (code == null) {
      code = Optional.empty();
    }
    return code;
  }
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/DatamartAppointment.java
```diff
@@ -0,0 +1,59 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class DatamartAppointment implements HasReplaceableId {
+  @Builder.Default private String objectType = "Appointment";
+
+  @Builder.Default private String objectVersion = "1";
+
+  private String cdwId;
+
+  private String serviceCategory;
```
jhulbert-va, line 25: We should check with Datasaurus to make sure these fields are required.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/AppointmentEntity.java
```diff
@@ -0,0 +1,61 @@
...
+@Table(name = "Appointment", schema = "app")
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class AppointmentEntity implements DatamartEntity {
+
+  @Id
+  @Column(name = "CDWId")
+  @EqualsAndHashCode.Include
+  private String cdwId;
```
jhulbert-va, line 32: This is now `BigInt` in the database. Should this be `BigInteger` instead of `String`?

ilaflamme-va: I was unsure on this

bschofield-va:
will reopen once entity and db schema are updated

PR #650, opened 2021-02-01
Api 2563 appointment entity and datamart object

ilaflamme-va:
relies on this [PR](https://github.com/department-of-veterans-affairs/lighthouse-starter-datamart/pull/3)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/AppointmentEntity.java
```diff
@@ -0,0 +1,72 @@
...
+@Builder
+@Table(name = "Appointment", schema = "app")
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@IdClass(CompositeCdwId.class)
+public class AppointmentEntity implements CompositeIdDatamartEntity {
+
+  @Transient
+  private final CompositeCdwId compositeCdwId =
```
bschofield-va, line 35: build this object on demand... not like this. This opens up all kind of state management issues. For example,
```
var a = AppointmentEntity.builder.cdwIdNumber(1).cdwIdResourceCode("a").build();
a.cdwIdResourceCode("b")
// a.compositeCdwId() is now hot garbage
```

bschofield-va: I don't mean to drone on, but this is the _two sources of truth_ problem. The _Highlander Rule_ applies here!

bschofield-va (approved):

PR #653, opened 2021-02-04
Api 4770 r4 appointment transformer

ilaflamme-va:
For the  `CodeableConcept`s we don't understand how to map to a code, we are only using `display` at this time.


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -0,0 +1,164 @@
...
+  List<Appointment.Participant> participants(List<DatamartReference> dmParticipants) {
+      return dmParticipants.stream().map(this::participant).collect(Collectors.toList());
+  }
+
+  Appointment.Participant participant(DatamartReference dmReference) {
+    if(Transformers.isBlank(dmReference)) {
+      return null;
+    }
+    // We only understand Appointment Participants that are Locations or Patients at this time.
+    if(dmReference.type().isPresent() && !(StringUtils.equals(dmReference.type().get(), "Location") || StringUtils.equals(dmReference.type().get(), "Patient"))) {
```
jhulbert-va, line 136: This feels odd. Essentially we're building an appointment where the only participant is a practitioner?

bschofield-va: This condition is hard to read and understand _intent_.  plz2rfactr

bschofield-va: Here are some hints to kick start you
```
private static final Set<String> SUPPORTED_PARTICIPANT_TYPES = Set.of("Patient","Location");
boolean isSupportedParticipant(DatamartReference r) {
  return !isBlank(r) 
    && r.hasTypeAndRefernce()
    && SUPPORTED_PARTICIPANT_TYPES.contains(r.get().type());
}
... meanwhile ...

Appointment.Participant participant(DatamartReference dmReference) {
  if (!isSupprtedParticipant(dmReference)) { return null; }
  return ... make a participant ...
}
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -0,0 +1,260 @@
...
+
+  String parseCdwIdResourceCode(String cdwId) {
+    var cdwIdParts = cdwId.split(":", -1);
+    if (cdwIdParts.length != 2) {
+      return null;
+    }
+    return cdwIdParts[1];
+  }
+
+  Appointment.AppointmentStatus parseTimes(Optional<Instant> start, Optional<Instant> end) {
```
bschofield-va, line 91: For what's it worth, this method isn't "parsing times". It's determining appointment status.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -0,0 +1,260 @@
...
+  }
+
+  Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
+    if (isBlank(maybeMinutesDuration)) {
+      return null;
+    }
+    return maybeMinutesDuration.get();
+  }
+
+  String parseCdwIdResourceCode(String cdwId) {
```
bschofield-va, line 83: This also isn't strictly "parsing cdw id resource code". I guess, in general, I don't like "parse" as a verb in a method since I find it ambiguous. It's not _english_ that I'd use to explain something to someone else.

```
// I would not say:
// set code to parse cdw id resource code cdwId
var code = parseCdwIdResourceCode(cdwId)

// I would say:
// set code to cdw resource code of cdw id
var code = cdwResourceCodeOf(cdwId)

// I would say:
// set code to cdw resource code from cdw id
var code = cdwResourceCodeFrom(cdwId)

```

ilaflamme-va: Ok, I see what you mean.

ilaflamme-va: Addressed

bschofield-va (approved):

PR #654, opened 2021-02-05
Api 4771 appointment controller and read

ilaflamme-va:


data-query/pom.xml
```diff
@@ -11,7 +11,7 @@
   <version>3.0.98-SNAPSHOT</version>
   <packaging>jar</packaging>
   <properties>
-    <datamart-starter.version>1.0.2</datamart-starter.version>
+    <datamart-starter.version>1.0.2-SNAPSHOT</datamart-starter.version>
```
jhulbert-va: Note To Self

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentController.java
```diff
@@ -0,0 +1,112 @@
...
+
+  private AppointmentRepository repository;
+
+  private LinkProperties linkProperties;
+
+  private VulcanConfiguration<AppointmentEntity> configuration() {
+    return VulcanConfiguration.forEntity(AppointmentEntity.class)
+        .paging(linkProperties.pagingConfiguration("Appointment", AppointmentEntity.naturalOrder()))
+        .mappings(Mappings.forEntity(AppointmentEntity.class).value("patient", "icn").get())
+        .rule(parametersNeverSpecifiedTogether("_id", "identifier", "patient"))
```
jhulbert-va, line 45: Why a rule for `_id` and `identifier` but no mappings?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/vulcanizer/VulcanizedReader.java
```diff
@@ -19,6 +20,11 @@
   VulcanizedTransformation<EntityT, DatamartT, ResourceT> transformation;
   /** The repository for the resource. */
   CrudRepository<EntityT, String> repository;
+  /**
+   * The compositeId repository for the resource. Use this instead of repository if the resource has
+   * composite ids.
+   */
+  CrudRepository<EntityT, CompositeCdwId> compositeIdRepository;
```
bschofield-va, line 27: this business of having two repos is still not acceptable

bschofield-va (changes requested):
Fix having two types of repos in Vulcanized components

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -162,15 +158,14 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
+    if (isBlank(compositeCdwId.cdwIdResourceCode())) {
       return null;
     }
-    String cdwIdResourceCode = cdwIdResourceCodeFrom(cdwId);
-    if (cdwIdResourceCode == null) {
+    if (allBlank(end, status)) {
       return null;
     }
-    if (cdwIdResourceCode.equals("W")) {
+    if (compositeCdwId.cdwIdResourceCode() == 'W') {
```
bschofield-va, line 168: Not a big fan of having logic and knowledge of the ID suffix `W` in two places. Seems better to ask the model, `isWaitList()` and `isAppointment()` vs. asking the model "does your ID end in W?"

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/vulcanizer/VulcanizedReader.java
```diff
@@ -28,22 +28,23 @@
...
+  public static <EntityT, DatamartT extends HasReplaceableId, ResourceT extends Resource, PkT>
+      VulcanizedReaderBuilder<EntityT, DatamartT, ResourceT, PkT> forTransformation(
           VulcanizedTransformation<EntityT, DatamartT, ResourceT> transformation) {
-    return VulcanizedReader.<EntityT, DatamartT, ResourceT>builder().transformation(transformation);
+    return VulcanizedReader.<EntityT, DatamartT, ResourceT, PkT>builder()
+        .transformation(transformation);
   }
 
-  private EntityT find(String publicId) {
+  private EntityT find(Function<String, PkT> toPrimaryKey, String publicId) {
```
bschofield-va, line 38: This function was supposed to be a _field_ not a method parameter.

ilaflamme-va: Oopsies. I fix

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/vulcanizer/VulcanizedReader.java
```diff
@@ -28,15 +28,25 @@
...
   }
 
   private EntityT find(String publicId) {
     String cdwId = transformation.witnessProtection().toCdwId(publicId);
-    Optional<EntityT> entity = repository.findById(cdwId);
+    PkT pk;
+    try {
+      pk = toPrimaryKey.apply(cdwId);
+    } catch (IllegalArgumentException e) {
+      throw new ResourceExceptions.NotFound(publicId);
```
bschofield-va, line 47: Help me understand why IllegalArgumentExceptions from all possible primary key transformers indicate the resource is Not Found. ^ Why only this exception type? This feels like very specific behavior.

bschofield-va: btw .. not saying this is wrong behavior, but it's not clear from this code _why_ this behavior exists. I think future us will need to understand this choice.

bschofield-va: also, there are no tests for this behavior

ilaflamme-va: For example, an unparsable id... `CompositeCdwId.fromCdwId(yucky-id)` would throw an IllegalArguementException.

This does feel _too specific_

ilaflamme-va: Any read on an ID that is garbage and can't be translated would always be Not Found in the DB.

bschofield-va: I think you can accomplish this by extracting this decision from the core processing.
```
... somewhere in Data Query ....
class CompositeCdwIds {
  public static String requireFromString(String cdwId) {
     try {
        return CompositeCdwId.fromCdwId(cdwId);
     } catch (IllegalArgumentException e) {
        throw ResourceExceptions.NotFound(cdwId);
     }
  }
}

... meanwhile ...

        .toPrimaryKey(CompositeCdwIds::requireFromString)
```

```

ilaflamme-va: Addressed.

bschofield-va (approved):
I'm good with java goodness

PR #662, opened 2021-02-09
Documentation on how a new resource should be added

bschofield-va:


how-add-a-new-resource.md
```diff
@@ -0,0 +1,52 @@
+# How to add a new resource
+
+### Goals
+
+- Learn about integration issues as soon as possible
+- Get a working steel thread complete, then add on functionality
+
+### Steps
+
+1. Discovery
```
jhulbert-va, line 10: These numbers be incrementing.

how-add-a-new-resource.md
```diff
@@ -0,0 +1,52 @@
...
+    1. Negotiate DB schema with CDW team.
+    1. Negotiate new resource scope with Auth team.
+1. Add datamart model support
+    1. Add Datamart model to DQ.
+    1. Update `DatamartValidationController` to support new type.
+    1. **Deploy** to give validation access to CDW team.
+1. Add synthetic data
+    1. Add JPA entity and repository to DQ.
+    1. Add test data to Synthetic DB.
+        1. **Deploy** Synthetic DB updates
```
jhulbert-va, line 22: I think this should be `3` not `2a`

how-add-a-new-resource.md
```diff
@@ -0,0 +1,52 @@
...
+    1. Update deployment ingress/alb definitions.
+    1. **Deploy** to allow testing CDW interaction.
+1. Add read support
+    1. Add ICN Majig to DQ.
+    1. Create Transformer
+    1. Add read support
+        1. Add ITs
+    1. **Deploy**
+1. Add initial search support
+    1. Add search by _id support
```
jhulbert-va, line 36: With Vulcan being as easy as it is, I think this should include `patient` (when appropriate) and `identifier` as well.

how-add-a-new-resource.md
```diff
@@ -0,0 +1,52 @@
...
+    1. Update resource controller to support parameter
+    1. Add ITs
+    1. **Deploy**
+1. Add machine-to-machine integration support
+    1. Update metadata
+    1. Add crawler support
+        1. Update `DataQueryScopes`
+    1. **Deploy**
+1. Publicize new resource
+    1. Update OpenAPI
```
jhulbert-va, line 50: Might be worth mentioning that this is done by updating the `DataQueryService` interfaces.

how-add-a-new-resource.md
```diff
@@ -0,0 +1,52 @@
...
+    1. **Deploy**
+    1. Update documentation prose
```
jhulbert-va, line 52: Feels like this should happen _before_ it gets updated in DQ.

how-add-a-new-resource.md
eclendenning-va: I think layering in on top of _just_ _id is still a good Idea here. That way we can test the steel thread and as we layer in we can decide what goes in together if the searches are simple enough.

how-add-a-new-resource.md
bschofield-va: markdown rendering will take of the numbers

how-add-a-new-resource.md
bschofield-va: ^ this is my intention here, get it out in QA connected to CDW running through kong etc. The point of this step is not to actually implement searches, but make sure the infrastructure is in place to invoke searches and produce bundles. There is no real search logic here.

how-add-a-new-resource.md
bschofield-va: So update the portal _before_ it's reflected in OpenAPI?

bschofield-va: FWIW, I think the OoenAPI should be done first. It's more detailed and low level. I'd hate to put something in the prose about a great new resource then it be completely absent from the swagger. That seems worse than stumbling on a great new resource in the swagger and wanting to know more.

jhulbert-va: What is your intention behind update documentation prose? My understanding is it would involve verifying the descriptions/mappings in fhir-resources, which should occur before we add it to the OpenAPI in DQ.

bschofield-va: The _prose_ is all of that header stuff on the dev portal that isn't Swagger.

bschofield-va: image
^ this stuff

bschofield-va: Perhaps this reads better now

how-add-a-new-resource.md
```diff
@@ -0,0 +1,52 @@
...
+- Get a working steel thread complete, then add on functionality
+
+### Steps
+
+1. Discovery
+    1. Document required data and searches. This will be used to negotiate table structure and datamart model.
+    1. Negotiate Datamart model with CDW team.
+    1. Negotiate DB schema with CDW team.
+    1. Negotiate new resource scope with Auth team.
+1. Add datamart model support
```
ilaflamme-va, line 15: Perhaps, a note that the discovery items _really_ need to be completed and finalized _before_ work begins, otherwise you're almost certainly going to have re-work to do.

bschofield-va: I think all of these steps are largely serial (why I used numbered list). While some things can be done in parallel, in general skipping a step increases chances of re-work.

eclendenning-va (approved):

jhulbert-va (approved):

ilaflamme-va (approved):

PR #665, opened 2021-02-11
API-5154: _id and identifier search support

jjabando1:


jhulbert-va (commented):
InvalidRequest Tests

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentController.java
```diff
@@ -44,13 +46,25 @@
...
+                .value("_id", this::loadCdwId)
+                .value("identifier", this::loadCdwId)
+                .value("patient", "icn")
                 .get())
+        .rule(parametersNeverSpecifiedTogether("_id", "identifier", "patient"))
         .defaultQuery(returnNothing())
         .build();
   }
 
+  private Map<String, ?> loadCdwId(String cdwId) {
```
jhulbert-va, line 59: Bearing in mind that we're moving most of our resources to the compositeID, should this be in a common class?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentControllerTest.java
```diff
@@ -143,6 +159,7 @@ void toBundle() {
   @SneakyThrows
   void validRequests(String query) {
     when(ids.register(any())).thenReturn(List.of(registration("1:A", "pa1")));
+    lenient().when(ids.lookup(any())).thenReturn(List.of(id("1:A")));
```
bschofield-va, line 162: why `lenient()`?  That's a bit of smell.

bschofield-va: Perhaps splitting this test into two types... one for IDs, one for not IDs.

bschofield-va: What happens when you give it an ID that it doesn't understand, e.g. `_id=nope`. I'd assume this would return an empty bundle.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentController.java
bschofield-va: I'm ok getting it right once here, then moving it when we need it the second time

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentControllerTest.java
jjabando1: Right it should just give back an empty bundle. In the IT it would be this check:
`test(200, Appointment.Bundle.class, r -> r.entry().isEmpty(), "Appointment?_id={id}", verifier.ids().unknown())`

jhulbert-va (approved):

bschofield-va (approved):

PR #668, opened 2021-02-12
Api 5152 appointment location search

ilaflamme-va:


bschofield-va (dismissed):

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -102,6 +103,13 @@ private static SystemDefinition local() {
         .build();
   }
 
+  private static TestIds.Appointments localAndLabAppointments() {
```
jhulbert-va, line 106: This isn't going to work. The encoding key is different between local and lab.

jhulbert-va (approved):

PR #671, opened 2021-02-12
API-5147: RawIT for R4 resources

jjabando1:


data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/RawIT.java
```diff
@@ -0,0 +1,152 @@
...
+  @SneakyThrows
+  public void assertFhirObject(String resourceName, String publicId) {
+    // Verify it is a raw response from the correct resource
+    String fhirObjectType =
+        readRaw(resourceName, publicId)
+            .jsonPath()
+            .using(JsonPathConfig.jsonPathConfig().charset("UTF-8"))
+            .get("objectType")
+            .toString();
+    assertThat(fhirObjectType).isEqualTo(resourceName);
```
jhulbert-va, line 48: This can just be 
```
assertFhirObject(resourceName, resourceName, publicId)
```

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/RawIT.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  @SneakyThrows
+  public void assertFhirObject(String resourceName, String publicId) {
+    // Verify it is a raw response from the correct resource
+    assertFhirObject(resourceName, resourceName, publicId);
+  }
+
+  @SneakyThrows
+  public void assertFhirObject(String oldResourceName, String r4ResourceName, String publicId) {
```
jhulbert-va, line 46: To make this more clear, I think `datamartResourceName` would be better than `oldResourceName`.

In addition, the signature seems a bit off. I think the first input should be the R4 resource and the second input should be the datamart resource. Thoughts?

ilaflamme-va:
Looks good to me.

ilaflamme-va:
As a note for the future, when you notice you're writing the same test multiple times, with different parameters:
`ParameterizedTest` is probably the way to go.

ilaflamme-va (approved):

jhulbert-va (approved):

PR #676, opened 2021-02-18
Api 5165 move date time parameter logic

eclendenning-va:
~update to non snapshot after https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/pull/119 merges~ :heavy_check_mark:

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/DateTimeParameters.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/DateTimeParameters.java
@@ -1,129 +1,19 @@
 package gov.va.api.health.dataquery.service.controller;
 
-import static com.google.common.base.Preconditions.checkArgument;
-
-import java.io.Serializable;
+import gov.va.api.health.fhir.api.FhirDateTimeParameters;
 import java.time.Instant;
-import java.time.OffsetDateTime;
-import java.time.ZoneId;
-import java.time.ZoneOffset;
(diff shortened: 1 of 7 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/DateTimeParameters.java
```diff
@@ -1,129 +1,19 @@
...
-public final class DateTimeParameters implements Serializable {
-  private static final int YEAR = 4;
-
-  private static final int YEAR_MONTH = 7;
-
-  private static final int YEAR_MONTH_DAY = 10;
-
-  private static final int TIME_ZONE = 20;
+@EqualsAndHashCode(callSuper = true)
+public class DateTimeParameters extends FhirDateTimeParameters {
```
jblefkove-va, line 13: Since this class is now just the database logic, how about `JpaDateTimeParameter`?

jblefkove-va, line 13: Not crazy about the inheritance. Would prefer `FhirDateTimeParameter` be its own immutable thing and for this class to have an internal `FhirDateTimeParameter` field.

bschofield-va: ^ 10/10 I prefer aggregation and composition over inheritance

bschofield-va (approved):

PR #679, opened 2021-03-09
API-5655 add facility and clinic fields for location

jblefkove-va:
Add `facilityId` and `locationIen` fields to Location datamart model

[see also](https://github.com/department-of-veterans-affairs/health-apis-datamart-synthetic-records/pull/105)


jhulbert-va (approved):
Looks good.

PR #682, opened 2021-03-11
API-5662 location clinic identifier

jblefkove-va:
Populate location identifier for clinic ID

```
$ curl -s 'http://localhost:8090/r4/Location/I2-7Q2IQADBCYM2BLJFCUUND7ITEE000000'
{
  "resourceType": "Location",
  "id": "I2-7Q2IQADBCYM2BLJFCUUND7ITEE000000",
  "identifier": [{
      "use": "usual",
      "type": {
        "coding": [{
            "system": "http://terminology.hl7.org/CodeSystem/v2-0203",
            "code": "FI",
            "display": "Facility ID"
          }
        ],
        "text": "Facility ID"
      },
      "system": "https://api.va.gov/services/fhir/v0/r4/NamingSystem/va-facility-identifier",
      "value": "vha_688"
    }, {
      "system": "https://api.va.gov/services/fhir/v0/r4/NamingSystem/va-clinic-identifier",
      "value": "vha_688_3485"
    }
  ],
  "status": "active",
  "name": "MENTAL HEALTH SERVICES",
  "description": "MENTAL HEALTH SERVICES",
  "mode": "instance",
  "type": [{
      "coding": [{
          "display": "PSYCHIATRY CLINIC"
        }
      ],
      "text": "PSYCHIATRY CLINIC"
    }
  ],
  "telecom": [{
      "system": "phone",
      "value": "202-745-8267"
    }
  ],
  "address": {
    "text": "50 IRVING STREET, NORTHWEST WASHINGTON DC 20422",
    "line": ["50 IRVING STREET, NORTHWEST"],
    "city": "WASHINGTON",
    "state": "DC",
    "postalCode": "20422"
  },
  "physicalType": {
    "coding": [{
        "display": "1ST FLOOR, 1D107"
      }
    ],
    "text": "1ST FLOOR, 1D107"
  },
  "managingOrganization": {
    "reference": "http://localhost:8090/r4/Organization/I2-INMEKVILJ2VASYKTSUKWKCF5ZI000000",
    "display": "WASHINGTON VA MEDICAL CENTER"
  }
}
```

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationTransformer.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationTransformer.java
@@ -58,17 +58,6 @@ public class R4LocationTransformer {
         .system("https://api.va.gov/services/fhir/v0/r4/NamingSystem/va-clinic-identifier")
         .value(facilityPrefix(facilityId) + facilityId.stationNumber() + "_" + clinicId)
         .build();
-    // .use(Identifier.IdentifierUse.usual)
-    // .type(
-    // CodeableConcept.builder()
-    // .coding(
-    // List.of(
-    // Coding.builder()
-    // .system("http://terminology.hl7.org/CodeSystem/v2-0203")
-    // .code("FI")
-    // .display("Facility ID")
-    // .build()))
-    // .build())
   }
 
   List<Identifier> identifiers(Optional<FacilityId> maybeFacilityId, Optional<String> maybeIen) {
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationTransformer.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationTransformer.java
@@ -1,8 +1,8 @@
 package gov.va.api.health.dataquery.service.controller.location;
 
+import static gov.va.api.health.dataquery.service.controller.FacilityTransformers.facilityIdentifier;
+import static gov.va.api.health.dataquery.service.controller.FacilityTransformers.fapiClinicId;
 import static gov.va.api.health.dataquery.service.controller.R4Transformers.asReference;
-import static gov.va.api.health.dataquery.service.controller.R4Transformers.facilityIdentifier;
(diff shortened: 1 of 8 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationTransformer.java
```diff
@@ -49,8 +50,21 @@ Address address(DatamartLocation.Address address) {
         .build();
   }
 
-  List<Identifier> identifiers(Optional<FacilityId> maybeFacilityId) {
-    return emptyToNull(asList(facilityIdentifier(maybeFacilityId.orElse(null))));
+  Identifier clinicIdentifier(FacilityId facilityId, String clinicId) {
+    if (facilityId == null || clinicId == null) {
```
bschofield-va, line 54: Do we need to consider blank values? Perhaps `isBlank(...)` instead of `== null` is safer?

jblefkove-va: good point

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationTransformer.java
```diff
@@ -49,8 +50,21 @@ Address address(DatamartLocation.Address address) {
...
+      return null;
+    }
+    return Identifier.builder()
+        .system("https://api.va.gov/services/fhir/v0/r4/NamingSystem/va-clinic-identifier")
+        .value(facilityPrefix(facilityId) + facilityId.stationNumber() + "_" + clinicId)
```
bschofield-va, line 59: This is not my favorite, but also not a hill I'm willing to die on.

I don't really like how `_` is used. It's part of the prefix so you get it for free, but you also have to add it explicitly before the clinic. I'd almost prefer this ID generate stuff be encapsulated vs doing it inline here and up ^^ there.

```
String asVapiFacilityId(FacilityId id) {
   return ... magic for "vha_123"
}

String asVapiClinicId(FacilityId id, String clinicId) {
   return ... magic for "vha_123_987"
}
```

bschofield-va (approved):

PR #685, opened 2021-03-30
API-5337 add DateUTC to appointment entity

that-guy-is-it:
This is part of a larger merge. It is done with this small change at first to allow for a Datamart update prior to the rest of the merges for DQ.


jblefkove-va:
`mitre.mv.db` doesn't have the `appointment.dateutc` column

jblefkove-va:
:shipit:

PR #697, opened 2021-04-15
API-5858: Fix R4 Metadata documentation

jabando-va:


jhulbert-va (commented):
> See VFQ metadata properties

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/MetadataProperties.java
```diff
@@ -66,4 +68,15 @@
...
   }
+
+  @Data
+  @Builder
+  @NoArgsConstructor
+  @AllArgsConstructor
+  @Accessors(fluent = false)
+  public static class VersionSpecificProperties {
+    private String id;
+    private String name;
```
jhulbert-va, line 79: With this boi, you can remove:
```
  private String id;
  private String dstu2Name;
  private String r4Name;
```  
and refactor the properties

bschofield-va (approved):

PR #702, opened 2021-04-21
API-4618 Remove Facility Identifier

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationTransformer.java
```diff
@@ -95,7 +72,7 @@ CodeableConcept physicalType(Optional<String> maybePhysicalType) {
   Location toFhir() {
     return Location.builder()
         .id(datamart.cdwId())
-        .identifier(identifiers(datamart.facilityId(), datamart.locationIen()))
+        //        .identifier(identifiers(datamart.facilityId(), datamart.locationIen()))
```
jblefkove-va, line 75: you don't want to delete this line... facility identifier should be removed, but we still want the clinic identifier

jblefkove-va (changes requested):
keep clinic identifier

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/location/R4LocationTransformerTest.java
```diff
@@ -54,22 +52,6 @@ void address() {
                 .build());
   }
 
-  @Test
-  void clinicIdentifier() {
```
jblefkove-va, line 58: this test should stay

jblefkove-va (approved):

PR #705, opened 2021-04-26
Updated to use new parent AccessTokens utility. Parent version needs to be de-SNAPSHOT'd

bschofield-va:


data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/AppointmentIT.java
```diff
@@ -109,7 +110,7 @@ void clinicianSearching() {
   @Test
   void oauthFlow() {
     assumeEnvironmentIn(Environment.STAGING_LAB, Environment.LAB);
-    String token = Oauth.exchangeToken(List.of("system/Appointment.read"));
+    String token = AccessTokens.get().forSystemScopes(List.of("system/Appointment.read"));
```
jhulbert-va, line 113: Uber nitpick. Why not use `forSystemScopes(String... scopes)`?

jhulbert-va (approved):

PR #711, opened 2021-04-30
API 4877 Location Searchable Clinic

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -100,6 +129,30 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+    return (token.hasSupportedSystem(CAPI_IDENTIFIER_SYSTEM) && token.hasExplicitCode())
+        || token.hasAnySystem();
+  }
+
+  private Specification<LocationEntity> tokenIdentifierSpecification(TokenParameter token) {
+    return token
+        .behavior()
+        .onExplicitSystemAndExplicitCode(
+            (system, code) -> {
+              var facilityId = selectFacilityId(code);
```
jblefkove-va, line 142: `clinicId`, not `facilityId`

jblefkove-va:
`curl -s 'http://localhost:8090/r4/Location?identifier=xyz'` should return an empty bundle, not 500

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -120,6 +120,9 @@ private static SystemDefinition local() {
         .addressCity("LYONS")
         .addressState("NJ")
         .addressPostalCode("07939")
+        .faciltyType("vha")
+        .locationIen("3485")
+        .stationNumber("688")
```
jblefkove-va, line 125: For organization, this is one field, `.facilityId("vha_442")`. Would `.clinicIdentifier("vha_688_3485")` be sufficient here?

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
```diff
@@ -33,13 +33,21 @@ void search() {
             "Location?_id={id}",
             testIds.unknown()),
         // Search by identifier
-        test(200, Location.Bundle.class, "Location?identifier={id}", testIds.location()),
```
jblefkove-va, line 36: are the existing tests going to be preserved?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -89,6 +101,22 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+      var locationIen = maybeClinicId.substring(maybeClinicId.lastIndexOf("_") + 1);
+      var facilityId = FacilityId.from(maybeClinicId.substring(0, maybeClinicId.lastIndexOf("_")));
+      Specification<LocationEntity> spec =
+          Specifications.<LocationEntity>select("facilityType", facilityId.type().toString())
+              .and(select("stationNumber", facilityId.stationNumber()));
+      if (spec == null) {
+        return null;
+      }
+      return spec.and(select("locationIen", locationIen));
+    } catch (IllegalArgumentException | StringIndexOutOfBoundsException e) {
```
jblefkove-va, line 115: Rather than catching `StringIndexOutOfBoundsException`, how about verifying `maybeClinicId.indexOf("_") >= 0`? If not, bail out and return null.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -89,6 +101,22 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
         .map(toBundle());
   }
 
+  private Specification<LocationEntity> selectClinicId(String maybeClinicId) {
+    try {
+      var locationIen = maybeClinicId.substring(maybeClinicId.lastIndexOf("_") + 1);
+      var facilityId = FacilityId.from(maybeClinicId.substring(0, maybeClinicId.lastIndexOf("_")));
```
jblefkove-va, line 107: based on how these vars are used, I think `facilityId` should be declared first

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -100,6 +128,30 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+    return (token.hasSupportedSystem(CAPI_IDENTIFIER_SYSTEM) && token.hasExplicitCode())
+        || token.hasAnySystem();
+  }
+
+  private Specification<LocationEntity> tokenIdentifierSpecification(TokenParameter token) {
+    return token
+        .behavior()
+        .onExplicitSystemAndExplicitCode(
+            (system, code) -> {
+              var clinicId = selectClinicId(code);
```
jblefkove-va, line 141: `clinicId` is confusing here... I think `clinicIdSpec` would be a better name? For that matter, I would rename `selectClinicId()` to `clinicIdSpec()` as well

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -100,6 +128,30 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+              if (clinicId == null) {
+                throw CircuitBreaker.noResultsWillBeFound("identifier", code, "Invalid facilityId");
```
jblefkove-va, line 143: clinic ID

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/location/R4LocationControllerTest.java
```diff
@@ -99,7 +113,7 @@ void readThrowsNotFoundWhenIdIsUnknown() {
   @ValueSource(
       strings = {
         "?_id=loc1",
-        "?identifier=loc1",
```
jblefkove-va, line 102: I think we want to keep `?identifier=loc1` too

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
that-guy-is-it: Now that we are working with empty bundles they can be preserved and at most reworked to expect empties.

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -120,6 +120,9 @@ private static SystemDefinition local() {
         .addressCity("LYONS")
         .addressState("NJ")
         .addressPostalCode("07939")
+        .faciltyType("vha")
```
jblefkove-va, line 123: `productionLocations()` will also need to be updated

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
```diff
@@ -40,6 +40,17 @@ void search() {
             b -> b.entry().isEmpty(),
             "Location?identifier={id}",
             testIds.unknown()),
+        test(
+            200,
+            Location.Bundle.class,
```
jblefkove-va, line 45: verify that the bundle is non-empty?

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
```diff
@@ -40,6 +40,17 @@ void search() {
...
+            "Location?identifier={identifier}",
+            testIds.locations().clinicIdentifier()),
```
jblefkove-va, line 47: Organization also has a `"Organization?identifier=https://api.va.gov/services/fhir/v0/r4/NamingSystem/va-facility-identifier|{facilityId}"` test, let's add something similar

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
```diff
@@ -40,6 +40,17 @@ void search() {
...
+        test(
+            200,
+            Location.Bundle.class,
+            b -> b.entry().isEmpty(),
+            "Location?identifier={identifier}",
+            "vha_123_4567"),
```
jblefkove-va, line 53: Since the other tests don't hardcode values, this should be something like `testIds.locations().unknownClinicIdentifier()`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -40,6 +46,9 @@
 @AllArgsConstructor(onConstructor = @__({@Autowired}))
 public class R4LocationController {
 
+  private static final String CAPI_IDENTIFIER_SYSTEM =
```
jblefkove-va, line 49: For organization, this is `FAPI_IDENTIFIER_SYSTEM` for Facilities API... I don't know that CAPI makes as much sense. How about `FAPI_CLINIC_IDENTIFIER_SYSTEM`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -40,6 +46,9 @@
 @AllArgsConstructor(onConstructor = @__({@Autowired}))
```
jblefkove-va, line 46: While we're in here, change the annotation to the more concise style

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
jblefkove-va: Also the unknown values for the other tests are all 5's

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -120,6 +120,10 @@ private static SystemDefinition local() {
         .addressCity("LYONS")
         .addressState("NJ")
         .addressPostalCode("07939")
+        .clinicIdentifier("vha_688_3485")
+        .fullClinicIdentifier(
+            "https://api.va.gov/services/fhir/v0/r4/NamingSystem/va-clinic-identifier|vha_688_3485")
+        .unknownClinicIdentifier("vha_123_4567")
```
jblefkove-va, line 126: to be consistent with other unknown values, this should be something like `vha_555_555`

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/TestIds.java
```diff
@@ -81,6 +81,9 @@
     @NonNull String addressCity;
     @NonNull String addressState;
     @NonNull String addressPostalCode;
+    @NonNull String clinicIdentifier;
+    @NonNull String fullClinicIdentifier;
```
jhulbert-va, line 85: I think my preference here would be include just the `clinicIdentifier` and then to hard code the system in the IT vs having two fields for the same identifier.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
```diff
@@ -40,6 +40,24 @@ void search() {
             b -> b.entry().isEmpty(),
             "Location?identifier={id}",
             testIds.unknown()),
+        test(
+            200,
+            Location.Bundle.class,
+            b -> !b.entry().isEmpty(),
```
jhulbert-va, line 46:
```
private Predicate<Location.Bundle> bundleIsNotEmpty() {
  return bundle -> !bundle.entry().isEmpty();
}
```

Then you can do things like:
```
bundleIsNotEmpty().negate()
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -89,6 +101,25 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+        return null;
+      }
+      var facilityId = FacilityId.from(maybeClinicId.substring(0, maybeClinicId.lastIndexOf("_")));
+      var locationIen = maybeClinicId.substring(maybeClinicId.lastIndexOf("_") + 1);
+      Specification<LocationEntity> spec =
+          Specifications.<LocationEntity>select("facilityType", facilityId.type().toString())
+              .and(select("stationNumber", facilityId.stationNumber()));
+      if (spec == null) {
+        return null;
+      }
```
jhulbert-va, line 116: Why this null check? Specifications are null safe, so why not just:
```
return           Specifications.<LocationEntity>select("facilityType", facilityId.type().toString())
              .and(select("stationNumber", facilityId.stationNumber()))
.and(select("locationIen", locationIen));
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -100,6 +131,30 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
         .build();
   }
 
+  private boolean tokenIdentifierIsSupported(TokenParameter token) {
+    return (token.hasSupportedSystem(FAPI_CLINIC_IDENTIFIER_SYSTEM) && token.hasExplicitCode())
```
jhulbert-va, line 135: This will only support `system|code` and `code`. I think we'll also want to support `system|`.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
jblefkove-va: spotbugs

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
jblefkove-va: For now it is sufficient to match Organization. There is a follow-up story to add `system|` query support to both resources: https://vajira.max.gov/browse/API-6794

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
jhulbert-va: It can, but why does that matter here? We perform null checks after using this method.
When I say null safe, I mean `Specification.and(null)` won't NPE.
I think we just need to null check the facilityId and locationIen (without both of those the request won't return anything anyway).

jblefkove-va: Adrian and I fought with this for a while; this null check is the simplest way to appease spotbugs.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
```diff
@@ -40,6 +46,24 @@ void search() {
...
```
jhulbert-va, line 52: You created the method, why no use it everywhere?

that-guy-is-it: 🤦 Missed the second one. Should I use it in every case, and remove the `negate()` as necessary?

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
```diff
@@ -4,63 +4,83 @@
...
-  TestIds testIds = DataQueryResourceVerifier.ids();
+  private final TestIds testIds = DataQueryResourceVerifier.ids();
+
+  private static Predicate<Bundle> bundleIsEmpty() {
+    return bundle -> bundle.entry().isEmpty();
+  }
+
+  private static Predicate<Bundle> bundleIsNotEmpty() {
+    return bundle -> !bundle.entry().isEmpty();
+  }
```
jhulbert-va, line 24: Why two methods doing opposites? You should be able to just `myPredicateMethod().negate()`

jblefkove-va: not a fan of `bundleIsNotEmpty().negate()` double-negative

jblefkove-va: Since these methods are mainly for readability (vs inlining the predicate), I'd rather have explicit `isEmpty` and `isNotEmpty`

jhulbert-va (approved):

jblefkove-va (approved):
:shipit:

PR #719, opened 2021-05-06
API 6503 Location Managing Organization Search

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/LocationEntity.java
```diff
@@ -56,6 +56,12 @@
   @Column(name = "LocationIen", nullable = true)
   private String locationIen;
 
+  @Column(name = "ManagingOrgIdNumber", nullable = true)
+  private Integer managingOrgIdNumber;
+
+  @Column(name = "ManagingOrgResourceCode", nullable = true)
+  private String managingOrgResourceCode;
```
jblefkove-va, line 63: char

PR #718, opened 2021-05-06
API-6945-update-token-mappings

jhulbert-va:
# [API-6945](https://vajira.max.gov/browse/API-6945)
In support of https://github.com/department-of-veterans-affairs/lighthouse-vulcan/pull/14 , this PR updates the deprecated tokenCsvList mappings in data-query.

PR #720, opened 2021-05-07
update entity and h2 with new columns

eclendenning-va:


jhulbert-va:
This boi will need to merge first: https://github.com/department-of-veterans-affairs/health-apis-datamart-synthetic-records/pull/129

jhulbert-va (dismissed):

jhulbert-va:
Merging this will also go unstable until https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/718 merges (vulcan will be auto-upgraded and some of the vulcan methods will have been deprecated).

jblefkove-va:
There will be a merge conflict on `mitre.mv.db` between this and https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/719

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionEntity.java
```diff
@@ -24,6 +25,8 @@
  * <pre>
  * CREATE TABLE [app].[Condition](
  *  [CDWId] [varchar](50) NOT NULL,
+ *  [CdwIdNumber] [bigint] NOT NULL,
+ *  [CdwIdResourceCode] [char](1) NULL,
```
jblefkove-va, line 29: I think this block comment is redundant and should be removed

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionEntity.java
```diff
@@ -45,6 +27,14 @@
 @EqualsAndHashCode(onlyExplicitlyIncluded = true)
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class ConditionEntity implements DatamartEntity {
+  @Column(name = "CdwIdNumber")
+  @EqualsAndHashCode.Include
```
bschofield-va, line 31: Should not be part of equals and hash code until these become the primary key.

jblefkove-va: It's permissible, but I think it makes more sense not to switch over hashcode and equals until these new columns are officially marked `@Id`

jhulbert-va (approved):

PR #721, opened 2021-05-07
MedicationRequest: Update patient and intent search capabilities

jhulbert-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/R4AllergyIntoleranceController.java
```diff
@@ -60,33 +59,17 @@
                 .reference(
                     "patient",
                     "icn",
-                    referencePatientSupportedResourceTypes(),
+                    Set.of("Patient"),
                     "Patient",
-                    this::referencePatientIsSupported,
-                    this::referencePatientValues)
+                    r -> R4Controllers.referencePatientIsSupported(r, linkProperties),
+                    R4Controllers::referencePatientValues)
```
bschofield-va, line 65:
```
Mappings.forEntity(...)
  .value(...)
  .add(PatientIdMapping.forLinks(linkProperties))
  .value(...)
  .get()
```
Can _all_ of the patient-to-icn column be standardized with a custom mapping?

PR #731, opened 2021-05-14
API-6605: Implement new appointment status logic using visitsid

apeirats-va:
Also see: https://github.com/department-of-veterans-affairs/health-apis-datamart-synthetic-records/pull/136

This PR implements logic described at bottom of the `visitSid` documentation: https://github.com/department-of-veterans-affairs/health-apis-data/blob/master/Documentation/Appt%20Status%20Logic%20Using%20VisitSID.md


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -212,7 +212,7 @@ Appointment toFhir() {
     return Appointment.builder()
         .resourceType("Appointment")
         .id(dm.cdwId())
-        .status(status(dm.start(), dm.end(), dm.status()))
+        .status(status(dm.start(), dm.status(), dm.visitSid(), Instant.now()))
```
bschofield-va, line 215: No need to pass `now` around. Just use create it where it is needed.

apeirats-va: This is meant to support unit tests, since the new logic is based on comparing the appt time to the current timestamp.

Alternatively, the Transformer could use a class-level Supplier, `Supplier<Instant> now = Instant::now;` which the unit test can override:
```
R4AppointmentTransformer.builder()
  ...
  .now(() -> Instant.parse(...))
  .build();
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -164,18 +164,18 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
    * </ul>
```
jblefkove-va, line 164: this javadoc should also be updated

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -188,21 +188,21 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
+  Appointment.AppointmentStatus statusFromStartAndVisitSid(
+      Optional<Instant> start, Optional<Long> visitSid, Instant now) {
+    if (start.isEmpty() || start.get().isAfter(now)) {
       return Appointment.AppointmentStatus.booked;
     }
-    if (start.isPresent() && end.isEmpty()) {
-      return Appointment.AppointmentStatus.arrived;
-    }
-    if (start.isPresent() && end.isPresent()) {
+    if (visitSid.isPresent()) {
```
jblefkove-va, line 202: If `visitSid` is not present, is that equivalent to `0`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -188,21 +188,21 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
         return null;
     }
   }
 
-  Appointment.AppointmentStatus statusFromStartAndEndTime(
-      Optional<Instant> start, Optional<Instant> end) {
-    if (start.isEmpty() && end.isEmpty()) {
+  Appointment.AppointmentStatus statusFromStartAndVisitSid(
+      Optional<Instant> start, Optional<Long> visitSid, Instant now) {
+    if (start.isEmpty() || start.get().isAfter(now)) {
```
jblefkove-va, line 199: This may not be a realistic case, but I think this should be `start.isPresent() && start.get().isAfter(now)`. e.g. if an appointment has not start time but does have `visitSid >0`, I would still expect that to be `fulfilled`.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -164,18 +164,18 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
       return null;
     }
     if (isWaitlist()) {
       return Appointment.AppointmentStatus.waitlist;
     }
     if (status.isEmpty()) {
-      return statusFromStartAndEndTime(start, end);
+      return statusFromStartAndVisitSid(start, visitSid, now);
     }
     switch (status.get()) {
```
jblefkove-va, line 180: could we do `status.orElse("NO ACTION TAKEN")` here to eliminate the above check?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
apeirats-va: I went with the logic described here: https://github.com/department-of-veterans-affairs/health-apis-data/blob/master/Documentation/Appt%20Status%20Logic%20Using%20VisitSID.md#proposed-additional-logic-using-appointmentdatetimeutc-and-visitsid

It's possible for a future appointment to have a `visitSid > 0`. It should be ignored according to the above, and always treated as `booked`. Should we revisit that?

jblefkove-va: I missed the note about future appointments having a visitSid. Leave it as is then.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
apeirats-va: It _should_ always be provided. It _should_ never be == 0 (_should_ always be either negative or positive).

If not present or null, I think it's indicative of an error and treated as such. I don't feel too strongly about it, though. I could treat it as a `-1`, along with a `log.warn(...)` message.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
apeirats-va: Going with a different approach- the test determines past/future times through `Instant.now().plus(Duration.ofDays(...))` instead of passing in explicit timestamps.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -156,28 +156,29 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
    * </ul>
    */
   Appointment.AppointmentStatus status(
-      Optional<Instant> start, Optional<Instant> end, Optional<String> status) {
+      Optional<Instant> start, Optional<String> status, Optional<Long> visitSid) {
     if (isBlank(compositeCdwId.cdwIdResourceCode())) {
       return null;
     }
-    if (allBlank(end, status)) {
+    if (allBlank(start, status)) {
```
jblefkove-va, line 172: should `visitSid` also be included?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -156,28 +156,29 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
     }
     if (isWaitlist()) {
       return Appointment.AppointmentStatus.waitlist;
     }
     if (status.isEmpty()) {
-      return statusFromStartAndEndTime(start, end);
+      return statusFromStartAndVisitSid(start, visitSid);
     }
-    switch (status.get()) {
+    switch (status.orElse("NO ACTION TAKEN")) {
```
jblefkove-va, line 181: with this defaulting you can remove the `status.isEmpty()` check above

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
apeirats-va: I think we can remove this check now that `status` defaults to `NO ACTION TAKEN`. Before, I just needed at least `status` or `start` to be provided. I think `visitSid` would've been useless on its own. 

By assuming a default value for `status` it removes the need for this check.

jblefkove-va:
Need to regenerate `mitre.mv.db` due to a merge... otherwise this looks ready to go

jblefkove-va (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
ilaflamme-va: Since we need the `visitSid` to make the correct decision for `Status`, and because `Status` is a required field, if it is missing, we should error.

ilaflamme-va:
Looks good to me.

ilaflamme-va (approved):

bschofield-va (changes requested):
just pausing this one minute

bschofield-va (approved):

PR #734, opened 2021-05-18
API 5008 Appointment Service Category Mapping

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -131,6 +131,47 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
         .collect(Collectors.toList());
   }
 
+  List<CodeableConcept> serviceCategory(Optional<String> maybeServiceCategory) {
+    if (isBlank(maybeServiceCategory)) {
+      return null;
+    }
+    var display = maybeServiceCategory.get();
+    String code;
+    switch (display) {
```
jblefkove-va, line 140: let's make this more readable by moving the switch into its own method

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -131,6 +131,47 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
+      case "SURGERY":
+        code = "S";
+        break;
+      default:
+        code = null;
+        break;
+    }
+    return List.of(
+        CodeableConcept.builder()
+            .coding(
```
jblefkove-va, line 165: Looking at how other `CodeableConcepts` are constructed in this file, I think you want to populate `text` as well

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
apeirats-va, line 140: I'd throw in a `toUppercase` and `trim` for an extra bit of safety.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -131,6 +131,47 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
+        break;
+      case "REHAB MEDICINE":
+        code = "R";
+        break;
+      case "SURGERY":
+        code = "S";
+        break;
+      default:
+        code = null;
+        break;
```
apeirats-va, line 161: If we can't map the value I think we should just return null here. No use returning a CodeableConcept with a null code and "gibberish" display.

apeirats-va: along with a warn log

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformerTest.java
```diff
@@ -26,7 +28,7 @@ void empty() {
             .compositeCdwId(CompositeCdwId.fromCdwId("1234:A"))
             .dm(DatamartAppointment.builder().build())
             .build();
-    assertThat(tx.toFhir()).isEqualTo(Appointment.builder().build());
+    assertThat(tx.toFhir()).isEqualTo(Appointment.builder().resourceType("Appointment").build());
```
jblefkove-va, line 31: merge conflict

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -217,14 +261,15 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
 
   Appointment toFhir() {
     return Appointment.builder()
+        .resourceType("Appointment")
```
jblefkove-va, line 264: merge conflict

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -217,14 +261,15 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
         .id(dm.cdwId())
-        .meta(meta(dm.lastUpdated()))
```
jblefkove-va, line 221: merge conflict

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -73,6 +74,34 @@ String description(Optional<String> maybeDescription) {
     return maybeDescription.get();
   }
 
+  String getServiceCategoryCode(String display) {
+    String code;
+    switch (display.toUpperCase().trim()) {
+      case "MEDICINE":
+        code = "M";
+        break;
```
jblefkove-va, line 82: now that the switch is its own method, you can return from each case instead of breaking

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -73,6 +74,34 @@ String description(Optional<String> maybeDescription) {
     return maybeDescription.get();
   }
 
+  String getServiceCategoryCode(String display) {
+    String code;
+    switch (display.toUpperCase().trim()) {
```
jblefkove-va, line 79: display may be null

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -73,6 +74,34 @@ String description(Optional<String> maybeDescription) {
     return maybeDescription.get();
   }
 
+  String getServiceCategoryCode(String display) {
```
jblefkove-va, line 77: The `get` prefix doesn't add anything, `serviceCategoryCode()` is sufficient

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
apeirats-va: I'd pass the Optional down here and null/isPresent check it here.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -140,6 +142,51 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
+                        .build()))
+            .text(display)
+            .build());
+  }
+
+  String serviceCategoryCode(Optional<String> displayOptional) {
+    if (isBlank(displayOptional)) {
+      return null;
+    }
+    switch (displayOptional.get().toUpperCase().trim()) {
```
jblefkove-va, line 172: `toUpperCase()` or `toLowerCase()` should specify Locale

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -140,6 +142,51 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
+                    Coding.builder()
+                        .system("http://www.va.gov/Terminology/VistADefinedTerms/44-9")
+                        .display(display)
+                        .code(code)
+                        .build()))
+            .text(display)
+            .build());
+  }
+
+  String serviceCategoryCode(Optional<String> displayOptional) {
```
jblefkove-va, line 168: this would be `maybeDisplay` to match the other method parameters

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -140,6 +142,51 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
   }
 
+  List<CodeableConcept> serviceCategory(Optional<String> maybeServiceCategory) {
+    if (isBlank(maybeServiceCategory)) {
+      return null;
+    }
+    String display = maybeServiceCategory.get();
+    String code = serviceCategoryCode(maybeServiceCategory);
+    if (code == null) {
+      log.warn("Appointment Service Category display value could not be mapped to a code.");
```
jblefkove-va, line 152: I'm not sold on this warning. I think if we have a non-blank display value but cannot map it to a code for whatever reason, we are still better off emitting a `CodeableConcept` with `system`, `display` and `text` set

apeirats-va: This may be my fault. I suggested the opposite of this earlier. I'm not sure what the right answer is so I'll defer to Peter.

Per story description: "If the value is not one of the 6 enumerated values then it is not emitted at all". Does that refer to the entire `serviceCategory` field, or just the `Coding.code` value?

jblefkove-va: My mistake. I didn't see that note in the story and it is pretty clear. (In general though we err on the side of being permissive and emitting the data we have, even if it is incomplete.)

apeirats-va: Then I'll suggest adding the display value to the warn log so it's easier to track in case this does pop up.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformerTest.java
jblefkove-va: you still need to revert this line

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -140,6 +143,52 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
 
+  List<CodeableConcept> serviceCategory(Optional<String> maybeServiceCategory) {
+    if (isBlank(maybeServiceCategory)) {
+      return null;
+    }
+    String display = maybeServiceCategory.get();
+    String code = serviceCategoryCode(maybeServiceCategory);
+    if (code == null) {
+      log.warn(
+          "Appointment Service Category display value {} could not be mapped to a code.", display);
```
jblefkove-va, line 154: On further thought, I think this log statement belongs in the `default` case of the switch

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -140,6 +143,54 @@ Integer minutesDuration(Optional<Integer> maybeMinutesDuration) {
...
+        return "0";
+      case "PSYCHIATRY":
+        return "P";
+      case "REHAB MEDICINE":
+        return "R";
+      case "SURGERY":
+        return "S";
+      default:
+        log.warn(
+            "Appointment Service Category display value {} could not be mapped to a code.",
```
bschofield-va, line 188: When you see this message in the log, how will investigate the record that caused it?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformerTest.java
```diff
@@ -133,6 +134,133 @@ void participants() {
...
+    assertThat(tx.serviceCategory(null)).isNull();
+    assertThat(
+            Appointment.builder()
+                .serviceCategory(tx.serviceCategory(Optional.of("SURGERY")))
+                .build())
+        .isEqualTo(
+            Appointment.builder()
+                .serviceCategory(
+                    List.of(
+                        CodeableConcept.builder()
```
bschofield-va, line 153: jeepers that's a lot of duplication... how about a utility method to makes this readable?
```....isEqualTo(appointmentWithServiceCategory("SURGERY","S"));```

bschofield-va: 6 * 14 = 84 lines of noise

jblefkove-va (approved):

PR #736, opened 2021-05-20
API 5007 Appointment ServiceType Mapping

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -273,6 +280,7 @@ Appointment toFhir() {
         .cancelationReason(cancelationReason(dm.cancelationReason()))
         .serviceCategory(serviceCategory(dm.serviceCategory()))
         .specialty(specialty(dm.specialty()))
+        .serviceType(serviceType(dm.serviceType()))
```
jblefkove-va, line 283: order of fields

jblefkove-va (approved):

PR #737, opened 2021-05-25
API 7244 Appointment Status Default

that-guy-is-it:
These small changes should minimize the number of `null` statuses in transformed Appointments.


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformerTest.java
```diff
@@ -6,6 +6,7 @@
 import gov.va.api.health.r4.api.datatypes.CodeableConcept;
 import gov.va.api.health.r4.api.datatypes.Coding;
 import gov.va.api.health.r4.api.resources.Appointment;
+import gov.va.api.health.r4.api.resources.Appointment.AppointmentStatus;
```
jblefkove-va, line 9: inline this

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -235,12 +235,15 @@ String serviceCategoryCode(Optional<String> maybeDisplay) {
     if (isBlank(compositeCdwId.cdwIdResourceCode())) {
```
jblefkove-va, line 235: `CompositeCdwId.fromCdwId(...)` cannot create an object where the resourceCode is null. I think we can remove this check.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformerTest.java
```diff
@@ -162,63 +277,21 @@ void serviceCategory() {
         .isEqualTo(serviceCategory("SURGERY", "S"));
   }
 
+  @ParameterizedTest
+  @MethodSource
+  void status(
+      Optional<Instant> startTime,
+      Optional<String> status,
+      Optional<Long> visitSid,
```
jblefkove-va, line 285: We can eliminate a lot of duplication in the parameters source method by making these arguments `Instant startTime`, `String status`, etc. Then just do `Optional.ofNullable()` for each one in here.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformerTest.java
```diff
@@ -162,63 +277,21 @@ void serviceCategory() {
...
+      Appointment.AppointmentStatus actual) {
```
jblefkove-va, line 286: This is the `expected` status; the actual is what `tx.status(...)` computes down in the test.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentTransformer.java
```diff
@@ -232,14 +232,14 @@ String serviceCategoryCode(Optional<String> maybeDisplay) {
...
+    if (isWaitlist()) {
+      return Appointment.AppointmentStatus.waitlist;
     }
     if (allBlank(start, status)) {
-      return null;
+      return Appointment.AppointmentStatus.booked;
     }
-    if (isWaitlist()) {
-      return Appointment.AppointmentStatus.waitlist;
+    if (allBlank(status, visitSid) && !isBlank(start)) {
```
jblefkove-va, line 241: Since we are defaulting to `booked` later, I don't think this check, or the `allBlank(start, status)` check above, are necessary.

Also, since status is a required field and we always want to populate it, the `default` case on the switch should also return `booked`.

jblefkove-va (approved):

PR #738, opened 2021-05-25
API-6794 FAPI system only queries

apeirats-va:
Adds system-only search support to R4 Organization. Previously, an organization search for an identifier token with a system id *only* resulted in an empty Bundle. It now returns a Bundle with any Organization with a non-empty identifier for the specified system.

~~It also adds a System-to-field-name helper class that generates necessary Functions for Vulcan's `TokenParameter` behavior callbacks.~~ (moved to Vulcan)


apeirats-va:
See branch in Vulcan for `Specifications.selectNotNull`: https://github.com/department-of-veterans-affairs/lighthouse-vulcan/tree/API-6794-specification-not-null

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/vulcanizer/SystemIdColumns.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import lombok.Data;
+import lombok.RequiredArgsConstructor;
+import org.springframework.data.jpa.domain.Specification;
+
+/** Provides Specification function shortcuts for simple system to field name mappings. */
+@RequiredArgsConstructor
+public class SystemIdColumns<EntityT> {
```
jblefkove-va, line 15: Should this class live in Vulcan?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/vulcanizer/SystemIdColumns.java
```diff
@@ -0,0 +1,107 @@
...
+      for (SystemColumnMapping<EntityT> column : columns) {
+        if (column.system().equals(system)) {
+          return column.withSystem().apply(system);
+        }
+      }
+      throw CircuitBreaker.noResultsWillBeFound(param, system, "Unknown system");
+    };
+  }
+
+  @Data
```
bschofield-va, line 64: This should `@Value` with `@Builder` and I'd like to see `@NonNull` to ensure that the object is create fully populated.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientController.java
```diff
@@ -143,7 +144,9 @@ boolean tokenGenderIsSupported(TokenParameter token) {
     return token
         .behavior()
         .onExplicitSystemAndExplicitCode(
-            (s, c) -> Specifications.<PatientEntityV2>select("gender", toCdwGender(c)))
+            SystemIdColumns.forEntity(PatientEntityV2.class, "gender")
```
bschofield-va, line 147: nit picking this a little... When first reading, I found this confusing. It's hard to tell what is string is what.

Perhaps restructuring could be more expressive, e.g.

```
SystemIdColumns.forEntity(PatientEntityV2.class)
  .requestParameter("gender")
  .addColumn("gender",PATIENT_GENDER_SYSTEM,this::toCdwGender)
  .matchSystemAndCode()
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/vulcanizer/SystemIdColumns.java
apeirats-va: Yes, I'm moving it over before I convert to PR

jblefkove-va (approved):

jblefkove-va (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -145,6 +146,8 @@ String readRaw(@PathVariable("publicId") String publicId, HttpServletResponse re
   private boolean tokenIdentifierIsSupported(TokenParameter token) {
     return (token.hasSupportedSystem(FacilityTransformers.FAPI_CLINIC_IDENTIFIER_SYSTEM)
             && token.hasExplicitCode())
+        || (token.hasSupportedSystem(FacilityTransformers.FAPI_CLINIC_IDENTIFIER_SYSTEM)
+            && token.hasAnyCode())
```
jblefkove-va, line 150: It occurs to me we have `(A && B) || (A && C)` here, which simplifies to `A && (B || C)`, i.e.

```
token.hasSupportedSystem(FacilityTransformers.FAPI_CLINIC_IDENTIFIER_SYSTEM && 
(token.hasExplicitCode() || token.hasAnyCode())
```

But in fact... the second clause can be eliminated entirely, because `(token.hasExplicitCode() || token.hasAnyCode())` is always true! `hasAnyCode()` is true if the token is `system|` form, while `hasExplicitCode()` is true in any other case (`system|code`, `|code`, or `code`)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -159,6 +162,11 @@ private boolean tokenIdentifierIsSupported(TokenParameter token) {
               }
               return clinicIdSpec;
             })
+        .onExplicitSystemAndAnyCode(
```
jblefkove-va, line 165: Should the `.onExplicitSystemAndExplicitCode()` case also be updated to match `R4OrganizationController`?

jblefkove-va:
Don't forget `Organization?identifier=I2-123` and `Location?identifier=I2-123`

jblefkove-va:
I took care of `Organization?identifier=I2-123` and `Location?identifier=I2-123`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
apeirats-va: Good catch.
This really should be just
```
return token.hasSupportedSystem(FacilityTransformers.FAPI_CLINIC_IDENTIFIER_SYSTEM)
        || (token.hasAnySystem() && token.hasExplicitCode());
```

apeirats-va: We don't want `hasAnySystem && hasAnyCode` right? That's just `param=|`. Although, looking at the Vulcan parser, it looks like that gets immediately rejected anyways.

So yes, that brings us down to `hasSupportedSystem(FAPI) || hasAnySystem`.

jblefkove-va (approved):

PR #740, opened 2021-05-28
API-7181-upgrade-dependencies

jhulbert-va:
# [API-7181](https://vajira.max.gov/browse/API-7181)

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/config/PathRewriteConfig.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/config/PathRewriteConfig.java
@@ -1,6 +1,7 @@
 package gov.va.api.health.dataquery.service.config;
 
 import gov.va.api.lighthouse.talos.PathRewriteFilter;
+import java.util.ArrayList;
 import java.util.Set;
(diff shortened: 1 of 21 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/config/PathRewriteConfig.java
```diff
@@ -22,7 +23,7 @@
     var registration = new FilterRegistrationBean<PathRewriteFilter>();
     registration.setOrder(Ordered.LOWEST_PRECEDENCE);
     PathRewriteFilter filter =
-        PathRewriteFilter.builder().removeLeadingPath(leadingPaths()).build();
+        PathRewriteFilter.builder().removeLeadingPath(new ArrayList<>(leadingPaths())).build();
```
bschofield-va, line 26: `List.of(...)`

bschofield-va: mvn... that's a `Set`, not a `List`

eclendenning-va (approved):

PR #743, opened 2021-06-04
API-7729 appointment sort by date

jblefkove-va:
https://vajira.max.gov/browse/API-7729

Add support for `Appointment?_sort=date`

[Vulcan PR](https://github.com/department-of-veterans-affairs/lighthouse-vulcan/pull/21)

[Swagger](https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/pull/127)

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/config/LinkProperties.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/config/LinkProperties.java
@@ -34,6 +34,10 @@ public class LinkProperties {
   private int defaultPageSize;
   private int maxPageSize;
 
+  public static Function<SortRequest, Sort> noSortableParameters() {
+    return r -> null;
+  }
+
   /**
    * Create standard page configuration for use for Vulcan based controllers. This is expecting a
    * resource name, e.g. DiagnosticReport and sorting, which is defined on the Datamart entities.
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/config/LinkProperties.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/config/LinkProperties.java
@@ -39,14 +39,14 @@ public class LinkProperties {
    * resource name, e.g. DiagnosticReport and sorting, which is defined on the Datamart entities.
    */
   public VulcanConfiguration.PagingConfiguration pagingConfiguration(
-      String resource, Sort sorting, Function<SortRequest, Sort> sortableParameters) {
+      String resource, Sort sortDefault, Function<SortRequest, Sort> sortableParameters) {
     return VulcanConfiguration.PagingConfiguration.builder()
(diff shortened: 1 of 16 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/config/LinkProperties.java
```diff
@@ -37,14 +38,16 @@
    * Create standard page configuration for use for Vulcan based controllers. This is expecting a
    * resource name, e.g. DiagnosticReport and sorting, which is defined on the Datamart entities.
    */
-  public PagingConfiguration pagingConfiguration(String resource, Sort sorting) {
-    return PagingConfiguration.builder()
+  public VulcanConfiguration.PagingConfiguration pagingConfiguration(
+      String resource, Sort sorting, Function<SortRequest, Sort> sortableParameters) {
```
jhulbert-va, line 42: Why not overload the method instead of making everything else pass null?

jblefkove-va: Overloading get messy if more parameters are added later; there are a lot of combinations. For one method vs two it's not a big deal, but the next time we added a parameter, the two methods would become four...

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/R4AllergyIntoleranceController.java
```diff
@@ -51,7 +51,7 @@
     return VulcanConfiguration.<AllergyIntoleranceEntity>forEntity(AllergyIntoleranceEntity.class)
         .paging(
             linkProperties.pagingConfiguration(
-                "AllergyIntolerance", AllergyIntoleranceEntity.naturalOrder()))
+                "AllergyIntolerance", AllergyIntoleranceEntity.naturalOrder(), r -> null))
```
bschofield-va, line 54: Can we turn `r -> null` into something a little more readable with a statically imported method, e.g. `noSortableParameters()`

data-query/src/main/java/gov/va/api/health/dataquery/service/config/LinkProperties.java
```diff
@@ -25,26 +27,31 @@
...
 public class LinkProperties {
-
   private String publicUrl;
   private String publicDstu2BasePath;
   private String publicStu3BasePath;
   private String publicR4BasePath;
   private int defaultPageSize;
   private int maxPageSize;
 
+  public static Function<SortRequest, Sort> noSortableParameters() {
```
bschofield-va, line 37: Should this have been in Vulcan's `PagingConfiguration`?

bschofield-va (dismissed):
I think this is fine, the only improvement would be to make `noSortableParameters()` reusable by moving to Vulcan.

jhulbert-va (dismissed):

jhulbert-va (approved):

PR #747, opened 2021-06-10
API 7734 Practitioner Role Create Datamart Entity

that-guy-is-it:
This PR includes a new Practitioner Role Entity, for use in Datamart population, and a Datamart Practitioner Role. Prior, the Datamart Practitioner Role was an inner class of Datamart Practitioner, and this is currently unchanged. However, because of the temporary duplication, an equally temporary testing class was added, which ensures the specified accessors work as expected. This is because it became necessary to either lower the testing threshold, add this class, or perform the full replacement now, of which this seemed like the best option. When the old Practitioner Role is removed, its tests should use the new implementation and this testing class should be removed as well.


that-guy-is-it:
This Draft PR is now relatively outdated due to `Datamart` related updates. It will be closed and reopened in a short time. The closure is due to potential refactors that may be slow and will unnecessarily clutter the PR.

PR #749, opened 2021-06-10
API-7908: practitioner basic name search

apeirats-va:
Adds basic `name`, `family`, and `given` search parameters to R4 Practitioner searches.

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentController.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentController.java
@@ -72,7 +72,6 @@ public class R4AppointmentController {
                 .value("location", "locationSid", this::publicIdToCdwIdNumber)
                 .values("_id", this::loadCdwId)
                 .values("identifier", this::loadCdwId)
-                .value("patient", "icn")
                 .get())
         .rules(List.of(parametersNeverSpecifiedTogether("_id", "identifier", "patient")))
         .defaultQuery(returnNothing())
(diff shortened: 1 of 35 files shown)


apeirats-va:
I noticed that Vulcan's rule configurations do not account for `string`'s `:exact` or `:contains` modifiers.

The following rule:
```
atLeastOneParameterOf("_id", "identifier", "name"),
```

disallows a request for `name:contains=...` or `name:exact=...` and forces a 400 response.

As it stands, the rule requires an explicit entry for each modifier, like so:

```
atLeastOneParameterOf("_id", "identifier", "name", "name:exact", "name:contains"),
```

Same goes for all other rules, like:
```
parametersNeverSpecifiedTogether("name", "given")
```


The other controllers don't account for this, so I left it as-is.

@bschofield-va , I think we should extend Vulcan's rules to account for string modifiers, rather than deal with all permutations at the controller level. Any opinions?

jhulbert-va (approved):

jhulbert-va (approved):

eclendenning-va (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/appointment/R4AppointmentController.java
```diff
@@ -72,7 +72,6 @@ private static Sort sortableParameters(SortRequest req) {
                 .value("location", "locationSid", this::publicIdToCdwIdNumber)
                 .values("_id", this::loadCdwId)
                 .values("identifier", this::loadCdwId)
-                .value("patient", "icn")
```
jblefkove-va, line 75: ?

PR #760, opened 2021-06-23
API-7560 use DatamartPractitionerRole for stu3

jblefkove-va:
Convert `Stu3PractitionerRoleController.java` to use `DatamartPractitionerRole` and `PractitionerRoleEntity`


data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -91,6 +91,7 @@ private static TestIds labIds() {
         .organizations(localAndLabOrganizations())
         .patient("1011537977V693883")
         .practitioner("I2-HRJI2MVST2IQSPR7U5SACWIWZA000000")
+        .practitionerRole("I2-6KYHN4VYERE5OHKPXWAPAU5BO4000000")
```
jblefkove-va, line 94: @bschofield-va is there any significance in using I2 vs I3 IDs here?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleControllerTest.java
```diff
@@ -156,9 +158,9 @@ void toBundle() {
         PractitionerRoleSamples.R4.asBundle(
             "http://fonzy.com/r4",
             List.of(
-                r4.practitionerRole(pubId1, orgPubId1, locPubId1),
-                r4.practitionerRole(pubId2, orgPubId2, locPubId2),
-                r4.practitionerRole(pubId3, orgPubId3, locPubId3)),
+                r4.practitionerRole(pubId1, pubId1, orgPubId1, locPubId1),
+                r4.practitionerRole(pubId2, pubId2, orgPubId2, locPubId2),
+                r4.practitionerRole(pubId3, pubId3, orgPubId3, locPubId3)),
```
apeirats-va, line 163: In my branch I created a separate set of `pracIdX` and `pracPubIdX` and passed those here.

jblefkove-va: Makes sense, we'll have to deal with that conflict in your branch; I went for the lowest impact change here since I didn't change R4.

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
bschofield-va: DQ has historically used I2 and should probably keep doing that. I3 are more compact, but otherwise offer no advantage.

apeirats-va (approved):

bschofield-va (approved):

PR #761, opened 2021-06-24
API-7560 DatamartPractitionerRole missing fields

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/DatamartPractitionerRole.java
```diff
@@ -158,4 +170,37 @@
...
+    public enum System {
+      phone,
+      fax,
+      pager,
+      email
+    }
+
+    public enum Use {
+      home,
+      mobile
```
nmoisan, line 203: also needs to include `work` as a possible value. I think I left that one out of the JSON documentation.

nmoisan:
LGTM other than the one comment on `use` enum

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/DatamartPractitionerRole.java
```diff
@@ -25,6 +25,8 @@
 
   private Optional<DatamartReference> practitioner;
 
+  private boolean active;
```
apeirats-va, line 28: `Boolean`?

jblefkove-va: This field is guaranteed non-null

jhulbert-va (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/DatamartPractitionerRole.java
apeirats-va: Will need to add them to synth dm

jblefkove-va: Good point, will go with `Boolean` for now then

apeirats-va (approved):

eclendenning-va (approved):

PR #763, opened 2021-06-25
API-3701: practitioner role search by practitioner identifer and name

apeirats-va:
Adds support to `R4PractitionerRoleController` for the search parameters:

- `practitioner.identifier` - allows for NPI or cdw id
- `practitioner.name`

### DIFF
--- a/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
+++ b/data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
@@ -20,11 +20,14 @@ public class PractitionerRoleIT {
   }
 
   @Test
-  void basic() {
+  void read() {
     verifier.verifyAll(
         test(200, PractitionerRole.class, "PractitionerRole/{id}", testIds.practitionerRole()),
-        test(404, OperationOutcome.class, "PractitionerRole/{id}", testIds.unknown()),
-        // search by _id
+        test(
+            404,
+            OperationOutcome.class,
+            "PractitionerRole/{id}",
+            testIds.unknown()), // search by _id
         test(
             200,
             PractitionerRole.Bundle.class,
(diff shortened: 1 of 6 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -55,12 +63,46 @@
...
         .mappings(
-            Mappings.forEntity(PractitionerRoleEntity.class).values("_id", this::loadCdwId).get())
+            Mappings.forEntity(PractitionerRoleEntity.class)
+                .values("_id", this::loadCdwId)
+                .tokens(
+                    "practitioner.identifier",
+                    this::tokenIdentifierIsSupported,
+                    this::tokenIdentifierSpecification)
+                .string("practitioner.given", "givenName")
+                .string("practitioner.family", "familyName")
```
jblefkove-va, line 73: I don't see explicit support required for `practitioner.given` or `practitioner.family` in [the spec](https://www.hl7.org/fhir/us/core/StructureDefinition-us-core-practitionerrole.html)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
```diff
@@ -55,12 +63,46 @@
...
+                parametersNeverSpecifiedTogether("practitioner.name", "practitioner.family")))
         .build();
   }
 
+  private Specification<PractitionerRoleEntity> identifierAnySystemAndExplicitCodeSpec(
+      String code) {
+    Specification<PractitionerRoleEntity> npiSpec =
+        Specifications.<PractitionerRoleEntity>select("npi", code);
+    try {
+      CompositeCdwId cdwId = CompositeCdwId.fromCdwId(witnessProtection.toCdwId(code));
```
jblefkove-va, line 96: You can use `CompositeCdwIds.optionalFromCdwId()` here to avoid dealing with `IllegalArgumentException`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/R4PractitionerRoleController.java
apeirats-va: Threw them in for sake of consistency with Practitioner.

jblefkove-va: Even if it's just two extra lines of Vulcan configuration, for now I would err on the side of omitting them.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
```diff
@@ -5,24 +5,130 @@
...
             PractitionerRole.Bundle.class,
             "PractitionerRole?_id={id}",
             testIds.practitionerRole()));
   }
+
+  @Test
+  void searchByName() {
+    // Grab subset of family/given names for startsWith tests
+    String startFamily = StringUtils.substring(testIds.practitioners().family(), 0, 3);
+    String startGiven = StringUtils.substring(testIds.practitioners().given(), 0, 3);
```
jblefkove-va, line 42: Is there significance to `StringUtils.substring()` here instead of plain `substring()`? Since this is test data, we can guarantee family and given will not be null.

apeirats-va: Null-safety is my only concern, although I realized we can guarantee the input to not be null. I've gotten used to preferring the StringUtils versions by default, so this is mostly just force of habit. I can refactor to use the native version if you prefer.

jblefkove-va: let's use the native version

jblefkove-va (approved):

PR #764, opened 2021-06-29
API-7562 use PractitionerRoleEntity for DSTU2 Practitioner

jblefkove-va:
Update `Dstu2PractitionerController` and `Dstu2PractitionerTransformer` to use `DatamartPractitionerRole` and stop using the embedded `DatamartPractitioner.PractitionerRole` class

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerController.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerController.java
@@ -85,11 +85,14 @@ public class Dstu2PractitionerController {
 
   List<PractitionerRoleEntity> findRolesById(String publicId) {
     String cdwId = witnessProtection.toCdwId(publicId);
-    if (!cdwId.endsWith(":S")) {
+    if (cdwId.length() <= 1 || cdwId.charAt(cdwId.length() - 2) != ':') {
       cdwId = cdwId + ":S";
     }
     return CompositeCdwIds.optionalFromCdwId(cdwId)
-        .map(id -> roleRepository.findByPractitionerIdNumber(id.cdwIdNumber()))
+        .map(
+            id ->
+                roleRepository.findByPractitionerIdNumberAndPractitionerResourceCode(
+                    id.cdwIdNumber(), id.cdwIdResourceCode()))
         .orElse(List.of());
   }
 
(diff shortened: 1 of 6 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerController.java
```diff
@@ -85,11 +83,23 @@ PractitionerEntity findById(String publicId) {
     return entity.orElseThrow(() -> new ResourceExceptions.NotFound(publicId));
   }
 
+  List<PractitionerRoleEntity> findRolesById(String publicId) {
+    String cdwId = witnessProtection.toCdwId(publicId);
+    if (!cdwId.endsWith(":S")) {
+      cdwId = cdwId + ":S";
+    }
+    return CompositeCdwIds.optionalFromCdwId(cdwId)
+        .map(id -> roleRepository.findByPractitionerIdNumber(id.cdwIdNumber()))
```
apeirats-va, line 92: No need to account for resource code on the search?

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerSamples.java
```diff
@@ -101,61 +99,6 @@ public DatamartPractitioner practitioner(String cdwId, String orgCdwId, String l
                       .build()))
           .gender(DatamartPractitioner.Gender.male)
           .birthDate(Optional.of(LocalDate.parse("1970-11-14")))
```
apeirats-va, line 101: The PractitionerRole object is still being served by both CDW and synth data payloads. Maybe we should keep it around until we work the tail end the migrations?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerController.java
jblefkove-va: Updated to include it, although it does seem superfluous since the assumption for the `:S` suffix is already baked into other places.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/PractitionerSamples.java
jblefkove-va: I've removed it from the unit test data to prove that all of the controllers/transformers are no longer using the field, but `DatamartPractitioner.practitionerRole` is still modeled and will remain so until the end of this effort.

apeirats-va (approved):

jhulbert-va (approved):

PR #766, opened 2021-07-02
API-8341: practitioner role specialty mapping entity

apeirats-va:
Follow up PR to synthetic datamart will use the entity to populate synthetic data.

### DIFF
--- a/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/PractitionerRoleSpecialtyMapEntity.java
+++ b/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/PractitionerRoleSpecialtyMapEntity.java
@@ -1,7 +1,5 @@
 package gov.va.api.health.dataquery.service.controller.practitionerrole;
 
-import gov.va.api.lighthouse.datamart.DatamartEntity;
-import gov.va.api.lighthouse.datamart.Payload;
 import java.io.Serializable;
 import java.math.BigInteger;
 import javax.persistence.Column;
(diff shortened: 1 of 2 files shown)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/PractitionerRoleSpecialtyMapEntity.java
```diff
@@ -0,0 +1,49 @@
...
+  @EqualsAndHashCode.Include
+  private char practitionerRoleResourceCode;
+
+  @Id
+  @Column(name = "SpecialtyCode")
+  @EqualsAndHashCode.Include
+  private String specialtyCode;
+
+  /** Primary Key. */
+  public SpecialtyMapCompositeId specialtyId() {
```
jhulbert-va, line 42: Not sure this will actually be used, I want to say remove it until you need it, but I'm not offended either way.

jhulbert-va (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitionerrole/PractitionerRoleSpecialtyMapEntity.java
apeirats-va: minimart wants it

apeirats-va: although I see your point that this logic could be pushed down to minimart

ilaflamme-va (approved):

PR #769, opened 2021-07-13
API 8040 Removing Practitioner Identifier System Only Query

that-guy-is-it:


data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerIT.java
```diff
@@ -82,10 +82,25 @@ void searchByIdentifier() {
...
-    // times out:
-    // test(200,Practitioner.Bundle.class,bundleHasResults(),"Practitioner?identifier=http://hl7.org/fhir/sid/us-npi|")
+            testIds.unknown()),
+        test(
+            200,
+            Practitioner.Bundle.class,
+            bundleHasResults(),
+            "Practitioner?identifier=|{npi}",
+            testIds.practitioners().npi()),
+        test(400, OperationOutcome.class, "Practitioner?identifier=|"),
```
jhulbert-va, line 92: Not sure this one needs to be tested in the ITs

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerControllerTest.java
```diff
@@ -158,7 +158,7 @@ void toBundle() {
       strings = {
         "?_id=111:S",
         "?identifier=111:S",
-        "?identifier=http://hl7.org/fhir/sid/us-npi|",
```
jhulbert-va, line 161: Is this not valid anymore? If not, it should be in an emptyBundle or invalidRequest test based on it's response.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerIT.java
```diff
@@ -82,10 +82,25 @@ void searchByIdentifier() {
             Practitioner.Bundle.class,
             bundleHasResults().negate(),
             "Practitioner?identifier=|{npi}",
-            testIds.unknown()));
-
-    // times out:
-    // test(200,Practitioner.Bundle.class,bundleHasResults(),"Practitioner?identifier=http://hl7.org/fhir/sid/us-npi|")
+            testIds.unknown()),
+        test(
```
jhulbert-va, line 86: These IT's don't need to be comprehensive for a token. There should really only be one test per parameter combination. All of these requests end up testing the same thing. Testing the token capabilities themselves should be done in the unit test level.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerControllerTest.java
that-guy-is-it: It is not, this story is about removing that capability. Will re-add it in an emptyBundle test.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerController.java
```diff
@@ -159,7 +159,10 @@ String readRaw(@PathVariable("publicId") String publicId, HttpServletResponse re
   }
 
   private boolean tokenIdentifierIsSupported(TokenParameter token) {
-    return token.hasSupportedSystem(PRACTITIONER_IDENTIFIER_SYSTEM_NPI) || token.hasAnySystem();
+    return (token.hasSupportedSystem(PRACTITIONER_IDENTIFIER_SYSTEM_NPI)
```
jhulbert-va, line 162: I think this can just return true?
Reading over it, it supports all token modes. 
At most, i'd say it just needs:
```
return token.system() == null || token.hasSupportedSystem(PRACTITIONER_IDENTIFIER_SYSTEM_NPI)
```
so that it kicks back any unknown systems as an empty bundle.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerControllerTest.java
jhulbert-va: Pretty sure that combo is still valid, so the empty bundle test should help you refine your `isSupported` method.

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerIT.java
```diff
@@ -82,10 +82,25 @@ void searchByIdentifier() {
...
+            200,
+            Practitioner.Bundle.class,
+            bundleHasResults(),
+            "Practitioner?identifier=|{npi}",
```
jblefkove-va, line 90: This is not correct. identifier=|{npi}` means that the system is explicitly blank while the code is `{npi}`, which should yield an empty bundle.

that-guy-is-it: ❗ Investigating.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerController.java
```diff
@@ -170,7 +173,7 @@ private boolean tokenIdentifierIsSupported(TokenParameter token) {
     return token
         .behavior()
         .onAnySystemAndExplicitCode(code -> identifierAnySystemAndExplicitCodeSpec(code))
-        .onExplicitSystemAndAnyCode(systemMappings.matchSystemOnly())
+        .onNoSystemAndExplicitCode(code -> identifierAnySystemAndExplicitCodeSpec(code))
```
jblefkove-va, line 176: The method names on this line are a clue that something is wrong.

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerControllerTest.java
apeirats-va: I think it's technically a valid request, so this should just return an empty bundle.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerController.java
```diff
@@ -159,7 +159,10 @@ String readRaw(@PathVariable("publicId") String publicId, HttpServletResponse re
   }
 
   private boolean tokenIdentifierIsSupported(TokenParameter token) {
-    return token.hasSupportedSystem(PRACTITIONER_IDENTIFIER_SYSTEM_NPI) || token.hasAnySystem();
+    return (token.hasSupportedSystem(PRACTITIONER_IDENTIFIER_SYSTEM_NPI)
+            || token.hasExplicitlyNoSystem()
```
apeirats-va, line 163: explicitly no system should not be supported

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerController.java
```diff
@@ -159,7 +159,8 @@ String readRaw(@PathVariable("publicId") String publicId, HttpServletResponse re
   }
 
   private boolean tokenIdentifierIsSupported(TokenParameter token) {
-    return token.hasSupportedSystem(PRACTITIONER_IDENTIFIER_SYSTEM_NPI) || token.hasAnySystem();
+    return (token.hasSupportedSystem(PRACTITIONER_IDENTIFIER_SYSTEM_NPI) || token.hasAnySystem())
+        && token.hasExplicitCode();
```
jblefkove-va, line 163: If `token.hasAnySystem()` is true, that implies `token.hasExplicitCode()` is true, i.e. any system implies you have a token like `|code` which must have an explicit code. So this could be simplified to:

```
return token.hasAnySystem()
	|| (token.hasSupportedSystem(PRACTITIONER_IDENTIFIER_SYSTEM_NPI)
		&& token.hasExplicitCode());
```

jblefkove-va (approved):

PR #774, opened 2021-08-12
API-9032: multiple codings for condition

jabando-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/Dstu2ConditionTransformer.java
```diff
@@ -110,6 +113,30 @@ CodeableConcept code(IcdCode icdCode) {
...
 
+  CodeableConcept code(DatamartCondition.SnomedCode snomedCode, DatamartCondition.IcdCode icdCode) {
+    if (snomedCode == null
+        || allBlank(snomedCode.code(), snomedCode.display())
+        || icdCode == null
+        || allBlank(icdCode.code(), icdCode.display())) {
+      return null;
+    }
+    return CodeableConcept.builder()
+        .text(snomedCode.display() + ", " + icdCode.display())
```
bschofield-va, line 124: This is incorrect. It should choose the text using the previous rules: SNOMED, ICD

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/Dstu2ConditionTransformer.java
```diff
@@ -110,6 +113,30 @@ CodeableConcept code(IcdCode icdCode) {
         .build();
   }
 
+  CodeableConcept code(DatamartCondition.SnomedCode snomedCode, DatamartCondition.IcdCode icdCode) {
```
bschofield-va, line 116: This all feels over complicated now:

_if I have two codes, use special method to add two codes. If I have one code, use special method to add one code_

How about: _if I have SNOMED code add it, if I have ICD code add it. If I added nothing, return nothing_

This allows you to consolidate the  code creating logic which is mostly duplicated and has flexibility for code 3 in the future. It also reduces the possible paths through the code to _one_ vs. two. Consider this: If we decided that http://snomed.info/sct was the incorrect system, we'd have to remember to change it _twice_. The duplicated nature makes it easy to miss and introduce Whack-a-mole bugs ... _I thought I fixed tha...t_

bschofield-va (changes requested):
Consolidate code creation to 1 path

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/Dstu2ConditionTransformer.java
```diff
@@ -25,17 +27,30 @@
   @NonNull private final DatamartCondition datamart;
 
   /**
-   * Return snomed code if available, otherwise icd code if available. However, null will be
+   * Return snomed code if available, icd code if available, or both. However, null will be
    * returned if neither are available.
    */
   CodeableConcept bestCode() {
+    List<CodeableConcept> codeList = new ArrayList<>();
```
bschofield-va, line 34: What is the maximum number of elements this array can have?
What is the amount of space this constructor is allocating?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/Dstu2ConditionTransformer.java
```diff
@@ -25,17 +27,30 @@
...
     if (datamart.hasSnomedCode()) {
-      return code(datamart.snomed().get());
+      codeList.add(code(datamart.snomed().get()));
```
bschofield-va, line 36: I don't really understand why you'd go through the trouble of creating up to _three_ CodeableConcept objects to _always_ return 1.

This kind of like wanting a sandwich, making one, wrapping it up in deli paper, immediately unwrapping it and throwing the paper away, then eating the sandwich. What was the point of paper?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/Dstu2ConditionTransformer.java
```diff
@@ -76,37 +62,45 @@ CodeableConcept category(Category category) {
...
+    CodeableConcept codeResult = null;
+    List<Coding> codingList = new ArrayList<>(2);
+    if (datamart.hasSnomedCode()) {
+      codingList.add(coding(datamart.snomed().get()));
+    }
+    if (datamart.hasIcdCode()) {
+      codingList.add(coding(datamart.icd().get()));
+    }
+    if (codingList.size() > 0) {
+      codeResult =
```
bschofield-va, line 79: in general, I think _return early_ is a better strategy than _return at end_. It generally requires less code and less nesting.

```
if (...) {
  return CodeableConcept.build()...build();
}
return null;
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/Dstu2ConditionTransformer.java
```diff
@@ -76,37 +62,45 @@ CodeableConcept category(Category category) {
...
+  CodeableConcept code() {
+    CodeableConcept codeResult = null;
+    List<Coding> codingList = new ArrayList<>(2);
+    if (datamart.hasSnomedCode()) {
+      codingList.add(coding(datamart.snomed().get()));
+    }
+    if (datamart.hasIcdCode()) {
+      codingList.add(coding(datamart.icd().get()));
+    }
+    if (codingList.size() > 0) {
```
bschofield-va, line 78: It's a little more elegant to use `xxx.isEmpty()` over `xxx.size() > 0`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/Dstu2ConditionTransformer.java
```diff
@@ -76,37 +62,45 @@ CodeableConcept category(Category category) {
...
 
-  CodeableConcept code(SnomedCode snomedCode) {
-    if (snomedCode == null || allBlank(snomedCode.code(), snomedCode.display())) {
+  /**
+   * Return snomed code if available, icd code if available, or both. However, null will be returned
+   * if neither are available.
+   */
+  CodeableConcept code() {
+    CodeableConcept codeResult = null;
+    List<Coding> codingList = new ArrayList<>(2);
```
bschofield-va, line 71: When naming things... don't repeat information. We _know_ this is a list _and_ the compiler guarantees it.
```
List<String> codingList
^__________________^
```

_List_ is superfluous... how about `codings`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/Dstu2ConditionTransformer.java
bschofield-va: One way to think about this, is _how can I write code that closely models how I'd explain it to another human_.

Would you say, "Hey Bry, if the coding list size is greater than 0, then we'll create codeable concept." or would you be more likely to say, "Hey Bry, if the coding list isn't empty, then we'll create codeable concept. And it will be dope" ?

Side note: kids say "dope" right?

bschofield-va (approved):

PR #777, opened 2021-08-19
Updated for two new columns in DiagnosticReportEntity.

taylorrassmannlibertyits:
Must be merged AFTER: https://github.com/department-of-veterans-affairs/health-apis-datamart-synthetic-records/pull/171


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/DiagnosticReportEntity.java
```diff
@@ -37,6 +38,12 @@
   @EqualsAndHashCode.Include
   private String cdwId;
 
+  @Column(name = "CdwIdNumber")
```
jhulbert-va, line 41: I believe this PR should also be updating from a single primary key to a composite primary key

jhulbert-va (approved):

bschofield-va:
Before we merge this... let's get lab's DB updated so we don't end up with a surprise should need to deploy DQ,

taylorrassmannlibertyits:
> Before we merge this... let's get lab's DB updated so we don't end up with a surprise should need to deploy DQ,

Deployed to lab: https://github.com/department-of-veterans-affairs/lighthouse-devops-support/issues/697

jblefkove-va (approved):

PR #782, opened 2021-09-14
API 9766 Immunization Updates Primary Source

that-guy-is-it:
Created a new field in `DatamartImmunization`, which will have to be merged first in order to unblock later parts of this task.


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
```diff
@@ -33,6 +33,8 @@
 
   private boolean wasNotGiven;
 
+  private boolean primarySource;
```
jblefkove-va, line 36: Needs to be non-primitive `Boolean` so that until the value is populated, it will default to `null` instead of `false`

that-guy-is-it: Ah, I had understood that it should be non-nullable now, so I used the primitive. Apologies. Fixing now.

jblefkove-va (approved):

PR #785, opened 2021-09-16
API-9760 replace DatamartAllergyIntolerance.status with clinicalStatus and verificationStatus

jblefkove-va:
https://vajira.max.gov/browse/API-9760

Update AllergyIntolerance transformers to use `DatamartAllergyIntolerance.clinicalStatus` and `DatamartAllergyIntolerance.verificationStatus` instead of `DatamartAllergyIntolerance.status`


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/allergyintolerance/Dstu2AllergyIntoleranceTransformer.java
```diff
@@ -92,21 +91,16 @@ Annotation notes(List<DatamartAllergyIntolerance.Note> notes) {
...
-    /*
-     * To correct a KBS compliance issue with Datamart data, we will report all 'confirmed' status
-     * values as `active`. From KBS ... "All allergy records should be marked active, excepting
-     * those with entered-in-error-flag set to True, which should be marked entered-in-error."
-     *
-     * Our analysis of the ETL process indicates 'confirmed' values may be returned.
-     */
-    if (status == Status.confirmed) {
-      return AllergyIntolerance.Status.active;
+    if (clinicalStatus != null) {
```
jhulbert-va, line 100: So if both are set we want to use clinical status?

jblefkove-va: If both are set `verificationStatus` takes precedence

jhulbert-va (approved):

PR #786, opened 2021-09-16
disable practitioner-role name search ITs

jblefkove-va:
Disable ITs that are timing out and disrupting deployments

[API-10007](https://vajira.max.gov/browse/API-10007) has been written to improve the performance of the PractitionerRole name search


data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/PractitionerRoleIT.java
```diff
@@ -48,6 +51,8 @@ void searchById() {
 
   @Test
   void searchByName() {
+    // These searches perform pooly in CDW environments
```
bschofield-va, line 54: I'd like to see a link to a story in JIRA that is turning this back on.

jhulbert-va (approved):

PR #794, opened 2021-09-24
API 4888 Immunization Composite CdwId

that-guy-is-it:
First of a series of PRs. This just ensures compatibility with future Composite CdwId updates in searches that do not use it, changing both DSTU2 and R4.


jblefkove-va (approved):
:shipit:

PR #797, opened 2021-09-28
API 4888 Immunization Entity Update

that-guy-is-it:
Update the Immunization Entity and a new mitre db for successful ITs.


jblefkove-va (commented):
:shipit:

jblefkove-va (approved):
:shipit:

PR #798, opened 2021-09-30
API 10292 Immunization Composite CdwId Searches

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
```diff
@@ -123,15 +139,19 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
       vulcanizedReader() {
     return VulcanizedReader
-        .<ImmunizationEntity, DatamartImmunization, Immunization, String>forTransformation(
+        .<ImmunizationEntity, DatamartImmunization, Immunization, CompositeCdwId>forTransformation(
             transformation())
         .repository(repository)
         .toPatientId(e -> Optional.of(e.icn()))
-        .toPrimaryKey(Function.identity())
-        .toPrimaryKeyFallback(cdwId -> cdwId.endsWith(":I") ? cdwId : cdwId + ":I")
+        .toPrimaryKey(CompositeCdwIds::requireCompositeIdStringFormat)
```
jblefkove-va, line 149: This `primaryKey` function is not compatible with supporting migration of suffixless Immunization IDs; if you pass `1234` into `requireCompositeIdStringFormat()`, it can't create a composite ID for that and thus throws an exception; the fallback function doesn't get a chance to run.

(You can verify this by reverting the changes to `SystemDefinitions` and running the Immunization ITs with the current suffixless IDs; they should still pass, but do not.)

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
```diff
@@ -123,15 +139,17 @@ public String readRaw(@PathVariable("publicId") String publicId, HttpServletResp
...
+        .<ImmunizationEntity, DatamartImmunization, Immunization, CompositeCdwId>forTransformation(
             transformation())
         .repository(repository)
         .toPatientId(e -> Optional.of(e.icn()))
-        .toPrimaryKey(Function.identity())
-        .toPrimaryKeyFallback(cdwId -> cdwId.endsWith(":I") ? cdwId : cdwId + ":I")
+        .toPrimaryKey(
+            cdwId ->
+                CompositeCdwIds.optionalFromCdwId(cdwId)
+                    .orElseGet(() -> CompositeCdwIds.requireCompositeIdStringFormat(cdwId + ":I")))
```
taylorrassmannlibertyits, line 152: Can this be simplified to: `CompositeCdwIds::requireCompositeIdStringFormat`?

taylorrassmannlibertyits: Answer: No it cannot :)

taylorrassmannlibertyits (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationController.java
jblefkove-va: why not `.orElse(requireCompositeIdStringFormat(...))` ?

jblefkove-va: I see it now

jblefkove-va (approved):
:shipit:

PR #799, opened 2021-10-06
asList updates for data query.

taylorrassmannlibertyits:
Special note: Since DQ is using `Arrays.AsList` and `singletonList` as null checks, not all instances of these could be taken out.


jblefkove-va:
Let's do the unit test updates in their own PR; those are low-risk and don't need to be reviewed as carefully.

PR #810, opened 2021-11-10
API-11214-pracititioner-and-practitioner-role-transformer-update

taylorrassmannlibertyits:
Updated transformers with new datamart fields for practitioner.


taylorrassmannlibertyits:
Hold off on merge until https://github.com/department-of-veterans-affairs/health-apis-datamart-synthetic-records/pull/213 is completed

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -88,6 +91,19 @@ static HumanName name(DatamartPractitioner.Name source) {
...
+    if (isBlank(maybePeriod)) {
+      return null;
+    }
+    DatamartPractitioner.Period period = maybePeriod.get();
+    String start = period.start().isEmpty() ? null : period.start().get().toString();
+    String end = period.end().isEmpty() ? null : period.end().get().toString();
+    if (allBlank(start, end)) {
+      return null;
+    }
+    return Period.builder().start(start).end(end).build();
```
jhulbert-va, line 104: `Optional` has a nice map function you could use which will only perform the action if the optional is present. I think it reads better than the ternary:
```
period.start().map(LocalDate::toString).orElse(null);
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -88,6 +91,17 @@ static HumanName name(DatamartPractitioner.Name source) {
...
   }
 
+  static Period period(Optional<DatamartPractitioner.Period> maybePeriod) {
+    if (isBlank(maybePeriod)) {
+      return null;
+    }
+
+    return Period.builder()
+        .start(maybePeriod.get().start().map(LocalDate::toString).orElse(null))
+        .end(maybePeriod.get().end().map(LocalDate::toString).orElse(null))
```
jhulbert-va, line 101: You'll still want to check that both start and end aren't blank.
```
   if (isBlank(maybePeriod) || allBlank(maybePeriod.get().start(), maybePeriod.get().end())) {
      return null;
    }
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -88,6 +91,17 @@ static HumanName name(DatamartPractitioner.Name source) {
     return List.of(source.get());
   }
 
+  static Period period(Optional<DatamartPractitioner.Period> maybePeriod) {
```
jhulbert-va, line 94: Just a note (no need to change it), `Optional.map` is also great to avoid optionals in your method signatures. e.g.
```
Line 54: 
    .period(address.period().map(this::period).orElse(null))

Line 94-97:
    Period period(DatamartPractitioner.Period period) {
        if (isBlank(period) || allBlank(period.start(), period.end())) {
          return null;
        }
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -107,12 +121,21 @@ static HumanName name(DatamartPractitioner.Name source) {
         .build();
   }
 
+  static List<Practitioner.Qualification> qualification(Optional<String> maybeQualificationText) {
```
jhulbert-va, line 124: We normally don't make theses methods static. It looks like they're only static for use in the tests. Instead, these can be non-static and the test class should have a `_transformer` method that creates a transformer that can then call the methods.
```
assertThat(_transformer().qualification(Optional.of("x"))).isEqualTo(expected);
```

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
taylorrassmannlibertyits: I like this way more, since I'm in the area still, I'll update to be cleaner for the future

jhulbert-va (approved):
LGTM

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerTransformer.java
```diff
@@ -81,6 +65,27 @@ static ContactPoint telecom(DatamartPractitioner.Telecom telecom) {
...
+  Address address(DatamartPractitioner.Address address) {
+    if (address == null
+        || allBlank(
+            address.line1(),
+            address.line2(),
+            address.line3(),
+            address.city(),
+            address.state(),
+            address.postalCode(),
+            address.period())) {
```
jblefkove-va, line 77: There is a subtle bug here. `address.period()` may be present with both its `start` and `end` null, so it will still map to a null FHIR period

jhulbert-va: Shouldn't the period method deal with this?

jblefkove-va: Right, if the only field present is a period with no `start` or `end`, we do not want to create an empty address

jblefkove-va (changes requested):
R4 only

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -146,10 +127,41 @@ static ContactPoint telecom(DatamartPractitioner.Telecom telecom) {
...
+  Address address(DatamartPractitioner.Address address) {
+    if (address == null
+        || allBlank(
+            address.line1(),
+            address.line2(),
+            address.line3(),
+            address.city(),
+            address.state(),
+            address.postalCode(),
+            address.period().map(this::period))) {
```
jhulbert-va, line 139: Not a fan of this. You'll end up doing the mapping twice if everything isn't blank.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Stu3PractitionerTransformer.java
```diff
@@ -104,10 +86,39 @@ static ContactPoint telecom(DatamartPractitioner.Telecom telecom) {
...
+        .state(address.state())
+        .postalCode(address.postalCode())
+        .build();
+  }
+
   private List<Address> addresses() {
     return emptyToNull(datamart.address().stream().map(adr -> address(adr)).collect(toList()));
   }
 
+  Period period(DatamartPractitioner.Period period) {
```
jblefkove-va, line 112: cruft

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerTransformer.java
```diff
@@ -81,6 +65,27 @@ static ContactPoint telecom(DatamartPractitioner.Telecom telecom) {
...
+  Address address(DatamartPractitioner.Address address) {
+    if (address == null
+        || allBlank(
+            address.line1(),
+            address.line2(),
+            address.line3(),
+            address.city(),
+            address.state(),
+            address.postalCode(),
+            address.period().map(this::period))) {
```
jblefkove-va, line 77: To Josh's point, let's split this check:
```
if (address == null) return null
period = address.period().map(...)
if (allBlank(..., period)) return null
```

jblefkove-va (approved):
:shipit:

jhulbert-va (approved):

PR #813, opened 2021-11-15
API 11228 Datamart Immunization Dose Quantity

that-guy-is-it:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
```diff
@@ -1,6 +1,7 @@
 package gov.va.api.health.dataquery.service.controller.immunization;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import gov.va.api.health.r4.api.datatypes.Quantity;
```
jblefkove-va, line 4: This needs to be a Datamart quantity, not FHIR; take a look at DatamartObservation

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
```diff
@@ -36,6 +37,8 @@
 
   private boolean primarySource;
 
+  private Optional<Quantity> doseQuantity;
```
jblefkove-va, line 40: will also need lazy getter

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
```diff
@@ -121,6 +131,29 @@ private void setEtlDate(String unused) {
...
+  public static final class Quantity implements AsList<Quantity> {
+    private Double value;
+
+    private String unit;
+
+    private String system;
+
+    private String code;
+
+    void setUnitCode(String code) {
```
jblefkove-va, line 148: These setters are for backwards compatibility; we don't need them here

jblefkove-va:
@department-of-veterans-affairs/lighthouse-datasaurus

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
```diff
@@ -121,6 +131,21 @@ private void setEtlDate(String unused) {
     data_absent_reason_unsupported
   }
 
+  /** Quantity. */
+  @Data
+  @Builder
+  @NoArgsConstructor(access = AccessLevel.PRIVATE)
+  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  public static final class Quantity implements AsList<Quantity> {
+    private Double value;
```
jblefkove-va, line 140: @nmoisan if Immunization `doseQuantity` is present, are any of these fields guaranteed non-null?

nmoisan: unfortunately no. there are cases where one or the other is still NULL. but we could force it such that we only send _usable_ data in the payload that has `dosage` and/or `doseUnit` populated. thoughts?

jblefkove-va: Do we do that kind of filtering elsewhere in datamart?

All of the quantity fields in the fhir spec are nullable, so it's not necessarily a problem. But it does indicate that everything should be `Optional<>` here

nmoisan: We do in some cases - but I think mostly in places where a sub-property is `required` if the optional parent is to be sent off. E.G. `Observation.Component.Code`

nmoisan: I'm just wondering how useful `units` is without `dosage`. But yeah Optional is probably safe

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/DatamartImmunization.java
```diff
@@ -121,6 +131,21 @@ private void setEtlDate(String unused) {
     data_absent_reason_unsupported
   }
 
+  /** Quantity. */
+  @Data
+  @Builder
+  @NoArgsConstructor(access = AccessLevel.PRIVATE)
+  @AllArgsConstructor(access = AccessLevel.PRIVATE)
+  public static final class Quantity implements AsList<Quantity> {
+    private Optional<Double> value;
```
jblefkove-va, line 140: Now that these are optional they also need lazy getters...

jblefkove-va:
Update `DatamartImmunizationTest` as well

jblefkove-va (approved):
:shipit:

PR #825, opened 2021-12-04
API-11719 add last-updated to datamart models

jblefkove-va:


jblefkove-va:
FYI @haidong @jtaylor-va @nmoisan

bschofield-va (approved):

PR #828, opened 2021-12-06
API 11591 Observation LastUpdated Record Search

that-guy-is-it:


data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -207,6 +207,7 @@ private static Observations observations() {
         .loinc2("777-3")
         .badLoinc("99999-9")
         .onDate("2015-04-15")
+        .lastUpdated("ge2008-01-01")
```
jblefkove-va, line 210: There is already `lastUpdated` defined for appointments. Adding a separate value for each resource does not seem like the correct approach; can we refactor the appointment `lastUpdated` to a shared location?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -257,4 +257,28 @@ void replaceReferences(Collection<DatamartObservation> resources) {
             .build(),
         entitiesPage);
   }
+
+  /** Search by Last Updated. */
+  @GetMapping(params = {"_lastupdated"})
```
jblefkove-va, line 262: shouldn't this be `_lastUpdated`?

that-guy-is-it: Yes, typo, sorry. Will fix on next push 👍

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -257,4 +257,28 @@ void replaceReferences(Collection<DatamartObservation> resources) {
...
+    ObservationRepository.PatientAndDateSpecification spec =
+        ObservationRepository.PatientAndDateSpecification.builder()
+            .patient(cdwPatient)
+            .dates(lastUpdated)
+            .build();
+    Page<ObservationEntity> entitiesPage = repository.findAll(spec, page(page, count));
+    return bundle(
+        Parameters.builder()
+            .add("patient", patient)
+            .addAll("_lastUpdated", lastUpdated)
```
taylorrassmannlibertyits, line 278: This makes the assumption that lastUpdated is not null. Is this handled by `@Valid`?

jblefkove-va: I think Spring handles it, we do the same thing for parameter `date` elsewhere in this file

jblefkove-va:
`Dstu2MetadataController` and `R4MetadataController` also need to be updated

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationTransformer.java
```diff
@@ -358,6 +368,7 @@ Observation toFhir() {
         Observation.builder()
             .resourceType("Observation")
```
jblefkove-va, line 369: while we're in here... remove this redundant `resourceType` here and in the r4 transformer

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -257,4 +257,28 @@ void replaceReferences(Collection<DatamartObservation> resources) {
...
+  public Observation.Bundle searchByPatientAndLastUpdated(
+      @RequestParam("patient") String patient,
+      @RequestParam("_lastUpdated") @Valid @DateTimeParameter @Size(max = 2) String[] lastUpdated,
+      @RequestParam(value = "page", defaultValue = "1") @Min(1) int page,
+      @CountParameter @Min(0) int count) {
+    String cdwPatient = witnessProtection.toCdwId(patient);
+    ObservationRepository.PatientAndDateSpecification spec =
+        ObservationRepository.PatientAndDateSpecification.builder()
+            .patient(cdwPatient)
+            .dates(lastUpdated)
```
jblefkove-va, line 272: this is searching on the wrong date

jblefkove-va (approved):
:shipit:

PR #833, opened 2021-12-16
API-11989 Add EncounterType to DatamartEncounter and tests and samples

taylorrassmannlibertyits:


jhulbert-va (approved):

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/encounter/EncounterSamples.java
```diff
@@ -40,6 +41,23 @@ public DatamartEncounter encounter(String cdwId, String patientIcn) {
...
+              EncounterType.builder()
+                  .coding(
+                      Optional.of(
+                          DatamartCoding.builder()
+                              .code(Optional.of("99212"))
+                              .display(Optional.of("OFFICE O/P EST SF 10-19 MIN"))
+                              .build()))
+                  .text(
+                      Optional.of(
+                          "OFFICE OR OTHER OUTPATIENT VISIT FOR THE EVALUATION AND MANAGEMENT OF AN "
```
jblefkove-va, line 54: For the purposes of test data, let's shorten this to something more reasonable

jtaylor-va:
Is there an test record for Inpatient Encounters?

jblefkove-va:
> Is there an test record for Inpatient Encounters?

For unit testing I think we only need one; we can make sure Inpatient and Outpatient Encounters are both represented in the sandbox data.

jblefkove-va (approved):

PR #834, opened 2021-12-17
Api 11909 encounter search and read

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-11909


data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4EncounterApi.java
```diff
@@ -0,0 +1,114 @@
...
+import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.enums.ParameterIn;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+
+public interface R4EncounterApi {
```
jblefkove-va, line 15: You need to add this to the parent openapi class

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/encounter/R4EncounterController.java
```diff
@@ -31,10 +44,48 @@
...
+  private Map<String, ?> loadCdwId(String publicId) {
+    String cdwId = witnessProtection.toCdwId(publicId);
+    try {
+      CompositeCdwId compositeCdwId = CompositeCdwId.fromCdwId(cdwId);
+      return Map.of(
+          "cdwIdNumber",
+          compositeCdwId.cdwIdNumber(),
+          "cdwIdResourceCode",
+          compositeCdwId.cdwIdResourceCode());
+    } catch (IllegalArgumentException e) {
```
jblefkove-va, line 78: You can use `optionalFromCdwId.optionalFromCdwId()` to avoid having to deal with this exception

jblefkove-va (approved):
:shipit:

PR #835, opened 2021-12-20
API 12037 observation procedue 500s

bschofield-va:
- Removed _lastUpdated reference from DSTU2 metadata controller
- Remove _lastUpdated from DSTU2 Observation


jblefkove-va (approved):
:shipit:

PR #836, opened 2021-12-20
API-11889: Observation referencerange null checks for quantity value/code

jabando-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/R4ObservationTransformer.java
```diff
@@ -321,9 +321,12 @@ static Quantity quantity(Optional<DatamartObservation.Quantity> maybeQuantity) {
   }
 
   private static SimpleQuantity simpleQuantity(Quantity quantity) {
-    if (quantity == null) {
+    if (quantity == null || quantity.value() == null) {
       return null;
     } else {
+      if (quantity.code() == null) {
```
jblefkove-va, line 327: If code is null, we don't want to populate system either, even if that's non-null?

jabando-va: Yeah, if I understood the story right: https://vajira.max.gov/browse/API-11889

jblefkove-va: Let's remove the above `else` to avoid nesting conditionals

bschofield-va (changes requested):
This is missing DSTU2, which is version we really need it in since it's what Apple is using and this issue was raised by Apple.

eclendenning-va (approved):

PR #838, opened 2021-12-21
API-11225 restore dstu2 procedure lastUpdated search

jblefkove-va:
Follow-up to https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/835


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/procedure/Dstu2ProcedureControllerTest.java
```diff
@@ -355,10 +355,40 @@ public void searchByPatientAndDateTwoDates() {
     }
   }
 
+  @Test
+  public void searchByPatientAndLastUpdated() {
+    Multimap<String, Procedure> proceduresByPatient = populateData();
+    assertThat(
+            json(
+                controller()
+                    .searchByPatient("p0", null, new String[] {"ge2018-01-01", "le2020"}, 1, 10)))
```
taylorrassmannlibertyits, line 364: Similar to line 321, should this new string array instead be `lastUpdated.getLeft/getRight()`?

jblefkove-va: The other test is verifying a bunch of different date combinations... for this one I'm okay with just hardcoding these two values

jhulbert-va (approved):

PR #841, opened 2021-12-23
API-11987 immunization vaccine-code R4 transformer

jblefkove-va:
https://vajira.max.gov/browse/API-11987

Update `R4ImmunizationTransformer` to transform full list of vaccine-codes


taylorrassmannlibertyits (dismissed):
:shipit:

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationTransformer.java
```diff
@@ -119,19 +123,23 @@ CodeableConcept vaccineCode(DatamartImmunization.VaccineCode vaccineCode) {
     if (vaccineCode == null || isBlank(vaccineCode.coding())) {
       return null;
     }
-    var coding = vaccineCode.coding().get(0);
-    if (allBlank(coding.code(), coding.display())) {
+    var codings =
+        emptyToNull(
+            Safe.stream(vaccineCode.coding()).map(c -> vaccineCodeCoding(c)).collect(toList()));
+    if (codings == null) {
```
jhulbert-va, line 129: I think we can just check that `size() == 0` here instead of iterating the whole list to determine it should be `null`.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/immunization/R4ImmunizationTransformer.java
```diff
@@ -119,19 +123,23 @@ CodeableConcept vaccineCode(DatamartImmunization.VaccineCode vaccineCode) {
     if (vaccineCode == null || isBlank(vaccineCode.coding())) {
       return null;
     }
-    var coding = vaccineCode.coding().get(0);
-    if (allBlank(coding.code(), coding.display())) {
+    var codings =
+        emptyToNull(
+            Safe.stream(vaccineCode.coding()).map(c -> vaccineCodeCoding(c)).collect(toList()));
```
jhulbert-va, line 128: `this::vaccineCodeCoding`

jhulbert-va, line 128: `vaccineCodingCode` can return `null` we should filter `Objects::nonNull`

jhulbert-va (approved):

PR #843, opened 2022-01-03
API-12186 add encounter appointment reference

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/encounter/DatamartEncounter.java
```diff
@@ -41,6 +41,8 @@
 
   private List<Participant> participants;
```
taylorrassmannlibertyits, line 43: Lazy initializer?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/encounter/R4EncounterController.java
```diff
@@ -121,7 +121,10 @@ String readRaw(@PathVariable("publicId") String publicId, HttpServletResponse re
         .replaceReferences(
             resource ->
                 Streams.concat(
```
jhulbert-va, line 123: Should we be filtering null references from the concated stream?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/encounter/DatamartEncounter.java
jhulbert-va: +1

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/encounter/R4EncounterController.java
jblefkove-va: `WitnessProtection` filters out the null references

taylorrassmannlibertyits (approved):
:shipit:

jhulbert-va (approved):

PR #846, opened 2022-01-04
API-12221 remove encounter from r4 openapi

jblefkove-va:
Remove encounter from openapi (and developer portal) until `Encounter.read` scope is available for consumers


taylorrassmannlibertyits (dismissed):
:shipit:

jhulbert-va (commented):
Is encounter in the metadata as well?

jhulbert-va:
> > Is encounter in the metadata as well?
> 
> yep

I think we should remove that as well.

jhulbert-va (approved):

PR #862, opened 2022-01-27
API-13054 Add icn field to DatamartPractitioner.

taylorrassmannlibertyits:


taylorrassmannlibertyits:
FYI @nmoisan @jtaylor-va @haidong

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/DatamartPractitioner.java
```diff
@@ -22,6 +23,9 @@
 
   @Builder.Default private int objectVersion = 1;
 
+  @JsonProperty("patientFullICN")
+  private String patientFullIcn;
```
jblefkove-va, line 27: Move this down with `npi` since it is another identifier

PR #864, opened 2022-01-28
API-13072 observation composite ID

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -222,7 +220,10 @@ void replaceReferences(Collection<DatamartObservation> resources) {
           firstIndex >= all.size()
               ? emptyList()
               : all.stream()
-                  .sorted(Comparator.comparing(ObservationEntity::cdwId))
+                  .sorted(
+                      Comparator.<ObservationEntity, BigInteger>comparing(
+                              ObservationEntity::cdwIdNumber)
+                          .thenComparingInt(ObservationEntity::cdwIdResourceCode))
```
jhulbert-va, line 226: Aren't these backwards? The number is the `int` and the code is a `char`.

jblefkove-va: `idNumber` is `BigInteger` and uses the generalized comparator function; `resourceCode` is `char`, which can be treated as `int` for sorting

taylorrassmannlibertyits (approved):
I'll miss the `fonz` in the unit tests.
:shipit:

PR #867, opened 2022-01-31
API-13056 Updated to Practitioner transformer to take icn.

taylorrassmannlibertyits:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/R4PractitionerTransformer.java
```diff
@@ -105,11 +109,31 @@ Address address(DatamartPractitioner.Address address) {
 
   List<Identifier> identifiers() {
     // ToDo is unknown the correct value to populate?
```
jblefkove-va, line 111: move this comment down too

jblefkove-va (approved):
:shipit:

PR #869, opened 2022-02-03
API-11952 Emit Encounter in R4Observation

taylorrassmannlibertyits:
Updated R4ObservationTransformer and samples for encounter references.
https://vajira.max.gov/browse/API-11952


jblefkove-va (approved):
:shipit:

PR #879, opened 2022-02-21
API-12834 Use state abbreviations in patient transformers.

taylorrassmannlibertyits:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
```diff
@@ -48,15 +48,15 @@ static Address address(DatamartPatient.Address address) {
...
+            address.stateAbbreviation(),
             address.postalCode(),
             address.country())) {
       return null;
     }
     return Address.builder()
         .line(emptyToNull(asList(address.street1(), address.street2(), address.street3())))
         .city(address.city())
-        .state(address.state())
+        .state(address.stateAbbreviation())
```
jblefkove-va, line 59: We should probably use `state` as a fallback if `stateAbbreviation` is blank

jblefkove-va:
Add a unit test for verifying abbreviation vs full state name

jblefkove-va (approved):
:shipit:

jblefkove-va (changes requested):
Approved, but let's wait until Feb 24 to merge

jblefkove-va (approved):
:shipit:

PR #882, opened 2022-03-02
API-13391 add DatamartDeviceRequest

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/devicerequest/DatamartDeviceRequest.java
```diff
@@ -0,0 +1,76 @@
...
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class DatamartDeviceRequest implements HasReplaceableId, AsList<DatamartDeviceRequest> {
+  @Builder.Default private String objectType = "DeviceRequest";
+
+  @Builder.Default private String objectVersion = "1";
+
+  private String cdwId;
+
+  private Optional<String> status;
```
jblefkove-va, line 25: `status` will be changed to an enum in a subsequent PR

haidong: In CDW, the `status` value comes from the `App.OrderStatus` table. It has 29 distinct values:
```
NULL
*Missing*
*Unknown at this time*
ACTIVE
ADM
CANCELLED
COMPLETE
DELAYED
DISCONTINUED
DISCONTINUED/EDIT
DO NOT USE
DUPLICATE
EXPIRED
file
FLAGGED
HOLD
ISSUED
LAPSED
MAKE APPOINTMENT
Manual D/C
NO SHOW
NO STATUS
PARTIAL RESULTS
PENDING
RECEIVED
RENEWED
SCHEDULED
UNRELEASED
ZZMAKE APPTMENT
```

jblefkove-va:
> The design is mostly settled. We are still working with KBS team to map the right OrderableItems, so we are only after CPAP/BIPAP machines. As part of discussion, things may change a bit. So be mindful that it may affect the API code.
> 
> We hard-code "order" for the `Intent` property. I notice that it's not defined. Would it be ok if I remove `Intent` from the payload?

That's fine. The exact payload details only impact the transformer logic; we can defer that and work on the rest of the pieces in the interim.

If `intent` will always be `order`, let's omit that from the payload and let the API hardcode it.

haidong (approved):

taylorrassmannlibertyits (approved):

PR #884, opened 2022-03-11
API-13624 add DeviceRequest entity and repository

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/devicerequest/DeviceRequestEntity.java
```diff
@@ -0,0 +1,79 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import org.springframework.data.domain.Sort;
+
+@Data
+@Entity
+@Builder
+@IdClass(CompositeCdwId.class)
+@Table(name = "DeviceRequest", schema = "app")
```
taylorrassmannlibertyits, line 29: Do you need `, indexes = @Index(columnList = "PatientFullICN")` for sorting purposes?

jblefkove-va: yep

taylorrassmannlibertyits (approved):

PR #890, opened 2022-03-17
API-13625 DeviceRequest read and search

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/devicerequest/DatamartDeviceRequest.java
```diff
@@ -22,6 +22,8 @@
 
   private String cdwId;
 
+  private Instant lastUpdated;
```
taylorrassmannlibertyits, line 25: Does there need to be a `mitre` update for this column being added??

jblefkove-va: The column is already present in the database

taylorrassmannlibertyits (approved):

PR #895, opened 2022-03-29
API-14852 DeviceRequest status update

taylorrassmannlibertyits:
Updated DeviceRequest status to use enum FHIR values. Updated transformer and test samples.


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/devicerequest/DatamartDeviceRequest.java
```diff
@@ -51,13 +52,25 @@
...
   /** Lazy initialization. */
-  public Optional<String> status() {
+  public Optional<Status> status() {
     if (status == null) {
       status = Optional.empty();
     }
     return status;
   }
 
+  public enum Status {
```
jblefkove-va, line 62: @haidong

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/devicerequest/R4DeviceRequestTransformer.java
```diff
@@ -31,10 +31,31 @@ Meta meta(Instant dmLastUpdated) {
     return Meta.builder().lastUpdated(lastUpdated).build();
   }
 
-  @SuppressWarnings("UnusedVariable")
-  private DeviceRequest.Status status(Optional<String> status) {
-    // status mapping TBD
-    return null;
+  private DeviceRequest.Status status(Optional<DatamartDeviceRequest.Status> status) {
```
jblefkove-va, line 34: I think you can use EnumSearcher for this?

PR #898, opened 2022-04-11
API-14903 Added display to datamart medication rxnorm and r4 transformer.

taylorrassmannlibertyits:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/R4MedicationTransformer.java
```diff
@@ -43,7 +43,7 @@ CodeableConcept bestCode() {
           .coding(
               Coding.builder()
                   .code(rxNorm.code())
-                  .display(rxNorm.text())
+                  .display(rxNorm.display() != null ? rxNorm.display() : rxNorm.text())
```
jblefkove-va, line 46: `isNotBlank()`

PR #900, opened 2022-04-14
API-15193 Updated parent versions to 10.0.0

taylorrassmannlibertyits:


jblefkove-va:
upgrade dependencies as well, and DQ itself goes to next major version

taylorrassmannlibertyits:
> upgrade dependencies as well

Which dependencies to upgrade?

jblefkove-va:
> > upgrade dependencies as well
> 
> Which dependencies to upgrade?

Any dependencies that also use health-apis-parent should be revved to the next major version, so that everything in the dependency tree is using the same version of Spring.

data-query-ids-mapping/pom.xml
```diff
@@ -4,11 +4,11 @@
...
-    <version>9.0.8</version>
+    <version>10.0.0</version>
     <relativePath/>
   </parent>
   <artifactId>data-query-ids-mapping</artifactId>
-  <version>4.0.86-SNAPSHOT</version>
+  <version>5.0.0-SNAPSHOT</version>
   <packaging>jar</packaging>
   <properties>
     <health-apis-ids.version>5.0.1</health-apis-ids.version>
```
jblefkove-va, line 14: update ids

data-query-tests/pom.xml
```diff
@@ -4,14 +4,14 @@
...
     <relativePath/>
   </parent>
   <artifactId>data-query-tests</artifactId>
-  <version>4.0.86-SNAPSHOT</version>
+  <version>5.0.0-SNAPSHOT</version>
   <packaging>jar</packaging>
   <properties>
-    <fhir-resources.version>10.0.27</fhir-resources.version>
+    <fhir-resources.version>11.0.0</fhir-resources.version>
     <health-apis-ids.version>5.0.0</health-apis-ids.version>
```
jblefkove-va, line 15: ditto

data-query/pom.xml
```diff
@@ -4,21 +4,21 @@
...
-    <version>9.0.8</version>
+    <version>10.0.0</version>
     <relativePath/>
   </parent>
   <artifactId>data-query</artifactId>
-  <version>4.0.86-SNAPSHOT</version>
+  <version>5.0.0-SNAPSHOT</version>
   <packaging>jar</packaging>
   <properties>
     <datamart-starter.version>4.0.0</datamart-starter.version>
```
jblefkove-va, line 14: update ids and datamart-starter

jblefkove-va (approved):

PR #902, opened 2022-04-18
API-13714 remove observation dstu2 query hack

jblefkove-va:


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationControllerTest.java
```diff
@@ -259,86 +234,64 @@ void searchByPatient() {
   }
 
   @Test
-  void searchByPatientAndCategoryAndOneDateHack() {
-    /*
```
taylorrassmannlibertyits, line 263: Why did you remove this comment block?

jblefkove-va: The query-hack-specific tests were removed

taylorrassmannlibertyits: Ok, but the other non hacks had a comment block too. I'm ok with the removal, just wanted to confirm.

jblefkove-va: It is not very clear in the diff, but `searchByPatientAndCategory()` (originally `searchByPatientAndCategoryNoHack()`) has no comment block, while `searchByPatientAndCategoryAndOneDate()` (originally `searchByPatientAndCategoryAndOneDateNoHack()`) does

taylorrassmannlibertyits (approved):

PR #903, opened 2022-04-20
API 13705 Device Natural Order

agarcia-va:
Straightforward update to add ICN-based sorting for `Device`. The existing Synthetic record schema had what was required.


taylorrassmannlibertyits:
> The existing Synthetic record schema had what was required.

I think this is partially true, I think you will need to update the jar in synth records once this is merged?

jblefkove-va:
Normally you need to update `mitre.mv.db` to include the patient index for the resource, but the local ITs passing regardless; it appears the order is coincidentally consistent. (There are only two devices in the test data.) So this can merge as-is.

jblefkove-va (approved):

PR #904, opened 2022-04-21
API-14989 Added reasonCode to DeviceRequest datamart and transformer.

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-14989


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/devicerequest/R4DeviceRequestTransformer.java
```diff
@@ -32,13 +38,33 @@ Meta meta(Instant dmLastUpdated) {
...
+                .code(reasonCode.code())
+                .display(reasonCode.display())
+                .build()
+                .asList())
+        .text(reasonCode.display())
+        .build();
+  }
+
+  List<CodeableConcept> reasonCodes(List<DatamartDeviceRequest.ReasonCode> reasonCodes) {
+    return emptyToNull(Safe.stream(reasonCodes).map(rc -> reasonCode(rc)).collect(toList()));
```
jblefkove-va, line 60: In one of the more recent java versions, you can replace `collect(toList())` with just `toList()`

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/devicerequest/R4DeviceRequestTransformer.java
```diff
@@ -32,13 +38,33 @@ Meta meta(Instant dmLastUpdated) {
...
 
+  CodeableConcept reasonCode(DatamartDeviceRequest.ReasonCode reasonCode) {
+    if (reasonCode == null
+        || allBlank(reasonCode.code(), reasonCode.codeVersion(), reasonCode.display())) {
+      return null;
+    }
+    return CodeableConcept.builder()
+        .coding(
+            Coding.builder()
+                .system("http://www.va.gov/Terminology/VistADefinedTerms/9000010-.07")
```
jblefkove-va, line 49: @nmoisan-va can you confirm this is the right system

nmoisan-va: 🤔 Is that what we have in Encounter and other places that use ICD codes?

nmoisan-va: I think it should be more similar to [Condition ](https://github.com/department-of-veterans-affairs/health-apis-data-query/blob/master/data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionTransformer.java#L133).

with the exception that will not ever provide a SNOMED code

jblefkove-va: It's the system currently used for Encounter `reasonCode`

taylorrassmannlibertyits: Updated to mirror Condition

jblefkove-va (approved):

PR #907, opened 2022-04-22
API-13597 observation sort

jblefkove-va:
Draft because this needs unit or integration tests


taylorrassmannlibertyits (approved):
:shipit:

PR #910, opened 2022-04-26
Api 13617 enable encounter sort

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-13617
and required
https://vajira.max.gov/browse/API-13707


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/encounter/R4EncounterSortTest.java
```diff
@@ -0,0 +1,137 @@
...
+    for (var dm : dms) {
+      repository.save(DM.entity(dm));
+    }
+  }
+
+  @Test
+  void sortByDate() {
+    var dm1 =
+        DM.encounter("1", "O", "6V6")
+            .period(
```
jblefkove-va, line 54: Instead of creating a new period, can you change the start time on the existing period? e.g. `DM.encounter().period().get().start(...)`

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/encounter/EncounterSamples.java
```diff
@@ -170,6 +173,22 @@ public EncounterEntity entity(
...
           .build();
     }
+
+    public EncounterEntity entity(DatamartEncounter dm) {
+      return EncounterEntity.builder()
+          .cdwIdNumber(CompositeCdwId.fromCdwId(dm.cdwId()).cdwIdNumber())
+          .cdwIdResourceCode(CompositeCdwId.fromCdwId(dm.cdwId()).cdwIdResourceCode())
+          .icn(dm.patient().reference().orElse(null))
+          .startDateTime(
+              Objects.requireNonNull(
```
jblefkove-va, line 183: Since we control the test data, you can safely assume here that period will be available, and dispense with `requireNonNull()`: `dm.period().get().start().get()...`

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/encounter/EncounterSamples.java
```diff
@@ -170,6 +173,22 @@ public EncounterEntity entity(
           .payload(json(dm))
           .build();
     }
+
+    public EncounterEntity entity(DatamartEncounter dm) {
```
jblefkove-va, line 177: With this method, do we still need the `entity(DatamartEncounter dm, String cdwIdNumber, String cdwIdResourceCode, String patientIcn)` method above?

taylorrassmannlibertyits: I considered taking it out but didn't want a huge ripple affect. I can update this

jblefkove-va: I'm less concerned about ripples when they're confined to tests

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/encounter/R4EncounterSortTest.java
jblefkove-va: This is what I was thinking:

```
var dm1 = DM.encounter("1", "O", "6V6");
dm1.period().get().start(Optional.of(LocalDate.of(2022, 12, 24)));
```

jblefkove-va (approved):

PR #919, opened 2022-05-10
Api 15714 r4 support observation type coding

taylorrassmannlibertyits:
remove whitespaces from type system


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/organization/R4OrganizationTransformer.java
```diff
@@ -96,12 +98,32 @@ static List<ContactPoint> telecoms(List<Telecom> maybeTelecoms) {
     return emptyToNull(maybeTelecoms.stream().map(tel -> telecom(tel)).collect(toList()));
   }
 
+  // todo remove this in the future once datamart has been updated to no longer have spaces in
```
jblefkove-va, line 101: Include the story number for removing this

taylorrassmannlibertyits: Is there one created?

jblefkove-va: No, please create one

PR #930, opened 2022-05-20
API-16161 DQ Removed all references and instances to bulk fhir

taylorrassmannlibertyits:


agarcia-va (approved):
LGTM! :shipit:!

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationorder/DatamartMedicationOrderTest.java
```diff
@@ -19,6 +19,42 @@ public void assertReadable(String json, DatamartMedicationOrder expected) {
     assertThat(dm).isEqualTo(expected);
   }
 
+  @Test
+  public void requiredFieldsOnlySample() {
+    assertReadable("datamart-medication-order-required-fields-only.json", requiredOnlySample());
```
jblefkove-va, line 24: What is the significance of this test?

taylorrassmannlibertyits: The was code coverage missing for `Optional<??>` values being null. This test goes through those fields.

jblefkove-va: In that case I'd prefer an explicit test for the lazy getters

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationorder/DatamartMedicationOrderTest.java
```diff
@@ -19,6 +20,24 @@ public void assertReadable(String json, DatamartMedicationOrder expected) {
     assertThat(dm).isEqualTo(expected);
   }
 
+  @Test
+  public void lazy() {
```
jblefkove-va, line 24: Remove public from these tests

data-query/src/test/java/gov/va/api/health/dataquery/service/controller/medicationorder/DatamartMedicationOrderTest.java
```diff
@@ -19,6 +20,24 @@ public void assertReadable(String json, DatamartMedicationOrder expected) {
...
+    DatamartMedicationOrder dm =
+        DatamartMedicationOrder.builder()
+            .dosageInstruction(
+                DatamartMedicationOrder.DosageInstruction.builder()
+                    .asNeeded(false)
+                    .build()
+                    .asList())
+            .build();
+    assertThat(dm.dosageInstruction().get(0).dosageText()).isEqualTo(empty());
```
jblefkove-va, line 33: This test is only checking `DosageInstruction`... let's check the lazy methods on the parent class, and other inner classes as well

jblefkove-va (approved):

PR #937, opened 2022-06-08
API-11219 DiagnosticReport lastUpdated search

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/R4DiagnosticReportTransformer.java
```diff
@@ -39,32 +39,33 @@ private List<CodeableConcept> category() {
         .asList();
   }
 
-  /** Determine the code value for the DiagnosticReport. */
   private CodeableConcept code() {
     return CodeableConcept.builder().text("panel").build();
   }
 
-  /** Determine the results value(s) for the DiagnosticReport. */
+  private Meta meta(Instant dmLastUpdated) {
```
taylorrassmannlibertyits, line 46: Not necessary right now, but as we keep adding more `lastUpdated` fields to our records, maybe we can have a generic `meta` method?

PR #940, opened 2022-06-16
API-11234 Updated R4Location to support lastUpdated.

taylorrassmannlibertyits:
https://vajira.max.gov/browse/API-11234


jblefkove-va:
1. Add an integration test
2. Update `R4MetadataController` (we neglected to do this for some previous resources, so backfill those as well)

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/LocationIT.java
```diff
@@ -80,6 +80,11 @@ void search() {
             "Location?identifier={unknown}",
             testIds.locations().clinicIdentifierUnknown()),
         test(200, Location.Bundle.class, "Location?name={name}", testIds.locations().name()),
+        test(
+            200,
+            Location.Bundle.class,
```
jblefkove-va, line 85: This test (and ultimately all `_lastUpdated` ITs) should verify that the bundle is non-empty

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/location/R4LocationController.java
```diff
@@ -140,6 +141,7 @@ private VulcanConfiguration<LocationEntity> configuration() {
                     "address-postalcode",
                     "_id",
                     "identifier",
+                    "_lastUpdated",
```
jblefkove-va, line 144: match this list order to the order above

jblefkove-va (approved):

PR #942, opened 2022-06-17
API-11231 Added lastUpdated seach for Patient.

taylorrassmannlibertyits:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/metadata/R4MetadataController.java
```diff
@@ -223,7 +223,7 @@ private List<CapabilityResource> resources() {
                 .build(),
             support("Patient")
                 .profileUrl("http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient")
-                .search(Set.of(SearchParam.ID, SearchParam.IDENTIFIER))
+                .search(Set.of(SearchParam.ID, SearchParam.IDENTIFIER, SearchParam.LAST_UPDATED))
```
jblefkove-va, line 226: Since we actually block patient searching at the gateway/kong level, I don't think we want to list this parameter here. (That is also why name, organization, etc. are omitted here.)

data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4PatientApi.java
```diff
@@ -153,6 +153,11 @@ Patient.Bundle patientSearch(
                       + " For Patients this identifier is an Integration Control Number (ICN)"
                       + " assigned by the Master Patient Index (MPI).")
           String identifier,
+      @Parameter(
+              in = ParameterIn.QUERY,
+              name = "_lastUpdated",
+              description = "The date when patient record was last updated.")
+          String lastUpdated,
       // Search by Name, Gender, and BirthDate are available but not as part of the patient flow
```
jblefkove-va, line 161: Based on this comment, we don't want to update the OpenAPI

jblefkove-va (approved):

PR #953, opened 2022-06-29
update r4 spec

bcastillomartel-va:
replaced the description of r4 openApi spec "Veteran" to "Patient".


bcastillomartel-va:
> 

Okay, I sent him a message

PR #958, opened 2022-07-02
include port in crawler base URL

jblefkove-va:


data-query-tests/crawler.log
```diff
@@ -0,0 +1,1647 @@
+Trusting host.docker.internal
```
taylorrassmannlibertyits, line 1: Do we need this file?

data-query-tests/src/main/docker/entrypoint.sh
```diff
@@ -120,15 +120,13 @@ doCrawlerTest() {
   [ -z "$DSTU2_CRAWL_PATH" ] && DSTU2_CRAWL_PATH=${INTERNAL_API_PATH}/dstu2
   [ -z "$R4_CRAWL_PATH" ] && R4_CRAWL_PATH=${INTERNAL_API_PATH}/r4
 
-  # Crawl DSTU2
   addToSystemProperties "crawler.url.replace" "${DATA_QUERY_REPLACEMENT_URL_PREFIX}/dstu2"
-  addToSystemProperties "crawler.base-url" "${DQ_URL}${DSTU2_CRAWL_PATH}"
+  addToSystemProperties "crawler.base-url" "${DQ_URL}:${DEPLOYMENT_TEST_PORT}${DSTU2_CRAWL_PATH}"
```
taylorrassmannlibertyits, line 124: Is DQ_URL unset now?

taylorrassmannlibertyits: Ah just a little bit lower I see where this is set. Please disregard

data-query-tests/crawler.log
jblefkove-va: We sure don't

data-query-tests/local-docker-image.sh
```diff
@@ -24,11 +22,11 @@ build() {
     -Ddocker.password="$DOCKER_PASSWORD"
 }
 
-
 run() {
   case $(uname) in
     Darwin) THIS_MACHINE="docker.for.mac.localhost";;
     Linux) THIS_MACHINE="localhost";;
+    MINGW*) THIS_MACHINE="host.docker.internal";;
```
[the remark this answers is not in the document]
jblefkove-va: It's case-sensitive; the OS on my machine (windows, git bash) is `MINGW64_NT-10.0-19042`

data-query-tests/local-docker-image.sh
```diff
@@ -37,19 +35,22 @@ run() {
     -e K8S_LOAD_BALANCER=$THIS_MACHINE \
     -e K8S_ENVIRONMENT=${ENV:-local} \
     -e SENTINEL_ENV=${ENV:-local} \
-    -e DQ_URL=http://$THIS_MACHINE:8090 \
```
jblefkove-va, line 40: Note that when `DQ_URL` is derived in `entrypoint.sh` now, it will be `http://$THIS_MACHINE`, without the port. The inclusion of the port here was actually a workaround for the bug of not including the port in `crawler.base-url`

taylorrassmannlibertyits (approved):

PR #970, opened 2022-07-27
add lastUpdated search support for medication resource

bcastillomartel-va:


jblefkove-va:
Reopen once R4 Medication is Vulcanized

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/MedicationIT.java
```diff
@@ -32,6 +32,13 @@ void search() {
             b -> isNotEmpty(b),
             "Medication?_id={id}",
             testIds.medication()),
+        test(
+            200,
+            Medication.Bundle.class,
+            b -> isNotEmpty(b),
+            "Medication?_id={id}&_lastUpdated={lastUpdated}",
```
jblefkove-va, line 39: Since `Medication` is not patient-centric, I should be able to search by `_lastUpdated` only.

bcastillomartel-va: what about this rule inside `MedicationController`:

 `atLeastOneParameterOf("_id", "identifier")`
 
 should I update the ruler to also have a `lastUpdated` entry ?

jblefkove-va: Yes

jblefkove-va (approved):

PR #986, opened 2022-09-09
github container registry

jblefkove-va:
https://github.com/department-of-veterans-affairs/health-apis-data-query-deployment/pull/157


bcastillomartel-va (dismissed):

KevinCranmer:
Do we want to update `entrypoint.sh` and `local-docker-image.sh` to be using the new docker name `ghcr.io/department-of-veterans-affairs/health-apis-data-query:latest` ?

jblefkove-va:
> Do we want to update `entrypoint.sh` and `local-docker-image.sh` to be using the new docker name `ghcr.io/department-of-veterans-affairs/health-apis-data-query:latest` ?

Good catch

bcastillomartel-va (approved):

PR #988, opened 2022-09-13
omg ... The AWS provided Local Dynamo DB is using 2014 binaries and can't work on ARM processors

bschofield-va:


jblefkove-va (approved):
:shipit:

PR #989, opened 2022-09-14
API-19742 restructure DatamartDocumentReference type

jblefkove-va:
see also https://github.com/department-of-veterans-affairs/health-apis-datamart-synthetic-records/pull/422

No backwards-compatibility since `DocumentReference` is not yet in production


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/documentreference/DatamartDocumentReference.java
```diff
@@ -1,8 +1,10 @@
 package gov.va.api.health.dataquery.service.controller.documentreference;
 
 import com.google.common.collect.Streams;
+import gov.va.api.health.dataquery.service.controller.observation.DatamartObservation.CodeableConcept;
```
jblefkove-va, line 4: whoops

bcastillomartel-va (approved):

PR #1000, opened 2022-09-27
add documentReference synthetic data

bcastillomartel-va:


data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
```diff
@@ -87,7 +87,7 @@ private static TestIds labIds() {
         .deviceRequest("I2-UFT532RWEIOS2WWZQWXUB7TKCT5SODPVTQJQD3ZNZUSAXXB4KZXQ0000")
         .diagnosticReport("I2-3ACWF6E3HPG6GLOSVWR2CIQNPI000000")
         .diagnosticReports(diagnosticReports())
-        .documentReference("I2-G3K4QNYOBIMT6FFGE4LEF7GSPDWK7QSFBDLW6ZNDXW6GO2NKPRQQ0000")
+        .documentReference("I2-3B2LP2KAI3TVQEB6EGKVZZKBVT5VOZPOMH2WZMEG3BUU7IDDRFYA0000")
```
KevinCranmer, line 90: The removed IDs exist but they exist for a new patient now. This caused the tests to fail with a 404. We were wondering why that is. Is it because ITs are set up for only the 883 patient?

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/DocumentReferenceIT.java
```diff
@@ -139,7 +139,7 @@ void searchByType_noSystemExplicitCode() {
             200,
             DocumentReference.Bundle.class,
             b -> isEmpty(b),
-            "DocumentReference?patient={patient}&type=|55107-7",
+            "DocumentReference?patient={patient}&type=|68629-5",
```
KevinCranmer, line 142: Why is this `b -> isEmpty(b)`? Same thing for Practitioner. Do we just not support searching for a type without including the system?

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
bcastillomartel-va: @jblefkove-va

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/DocumentReferenceIT.java
bcastillomartel-va: @jblefkove-va

data-query-tests/src/main/java/gov/va/api/health/dataquery/tests/SystemDefinitions.java
jblefkove-va: Only certain patients are included in `mitre.mv.db`:

https://github.com/department-of-veterans-affairs/health-apis-datamart-synthetic-records/blob/qa/minimart-manager/export-h2.sh#L28

The ITs for the patient-centric references are set up around `1011537977V693883`, because that is the required patient for the static test token. (Once ITs no longer need to run in K8S, we will have more flexibility for testing other patients.)

data-query-tests/src/test/java/gov/va/api/health/dataquery/tests/r4/DocumentReferenceIT.java
jblefkove-va: https://www.hl7.org/fhir/r4/search.html#token

`type=|68629-5` means that the code is `68629-5` with explicitly no system set; since our system for this field is always `http://loinc.org`, this query will always return no results

jblefkove-va (approved):

PR #1002, opened 2022-09-30
update Procedure with DateUTC

bcastillomartel-va:


data-query/src/test/java/gov/va/api/health/dataquery/service/controller/documentreference/R4DocumentReferenceTransformerTest.java
```diff
@@ -85,6 +87,13 @@ void content() {
                         .build())
                 .build()
                 .asList());
+    assertThat(tx.attachment(null)).isNull();
+    var emptyAttachment = DatamartDocumentReference.Attachment.builder().build();
+    assertThat(tx.attachment(emptyAttachment)).isNull();
+    assertThrows(NullPointerException.class, () -> tx.attachmentReference(null).equals(null));
```
jblefkove-va, line 93: wouldn't this just be `assertThat(tx.attachmentReference(null)).isNull()`?

bcastillomartel-va: that would be correct

jblefkove-va (approved):

PR #1005, opened 2022-10-11
Api 20610 create custom sql server dialect

bcastillomartel-va:


data-query/src/main/resources/application.properties
```diff
@@ -47,6 +47,7 @@ spring.datasource.hikari.minimum-idle=2
 spring.datasource.hikari.maximum-pool-size=20
 spring.jpa.hibernate.naming.physical-strategy=org.hibernate.boot.model.naming.PhysicalNamingStrategyStandardImpl
 spring.jpa.properties.hibernate.globally_quoted_identifiers=true
+spring.jpa.properties.hibernate.dialect=gov.va.api.health.dataquery.service.DangerZoneNoLockCustomDialect
```
bschofield-va, line 50: You can't/shouldn't set the default dialect to be SqlServer ... you have no control on what type of database is going to be used ,e.g. H2 for integration tests. You also don't know if such a dangerous dialect is appropriate for all deployments/usages, e.g. Synthetic Data updates should not be using `nolock`.

data-query/src/main/java/gov/va/api/health/dataquery/service/DangerZoneNoLockCustomDialect.java
```diff
@@ -0,0 +1,12 @@
+package gov.va.api.health.dataquery.service;
+
+import org.hibernate.LockOptions;
+import org.hibernate.dialect.SQLServer2016Dialect;
+
+public class DangerZoneNoLockCustomDialect extends SQLServer2016Dialect {
+
+  @Override
+  public String appendLockHint(LockOptions lockOptions, String tableName) {
+    return tableName + " with (nolock)";
```
bschofield-va, line 10: This is incorrect. You are completely discounting any other locking behavior that might be desired for a particular query. 

You want to _default_ to `nolock` but you don't want to force `nolock` and forbid all other lock hints. You need to invoke the super implementation at some point... e.g. 
```
if lock options is NONE
then append "with (nolock)"
otherwise use default sqlserver dialect behavior
```

data-query/src/test/java/gov/va/api/health/dataquery/service/DangerZoneNoLockCustomDialectTest.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.hibernate.LockOptions;
+import org.junit.jupiter.api.Test;
+
+class DangerZoneNoLockCustomDialectTest {
+
+  DangerZoneNoLockCustomDialect dangerZoneNoLockCustomDialect = new DangerZoneNoLockCustomDialect();
+
+  @Test
+  void appendLockHintTest() {
```
bschofield-va, line 13: Where is the opposite of this test?
Please test that accounts for when your `if` is false.

data-query/src/main/java/gov/va/api/health/dataquery/service/DangerZoneNoLockCustomDialect.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.Generated;
+import org.hibernate.LockOptions;
+import org.hibernate.dialect.SQLServer2016Dialect;
+
+public class DangerZoneNoLockCustomDialect extends SQLServer2016Dialect {
+  @Override
+  public String appendLockHint(LockOptions lockOptions, String tableName) {
+    String tableNameWithLock = super.appendLockHint(lockOptions, tableName);
+    if (tableNameWithLock.equals(tableName)) {
+      tableNameWithLock += " with (nolock)";
```
jblefkove-va, line 13: Let's  invert this check so that this becomes a guard clause instead, i.e. if the super method result does not match the original table name, return it, otherwise continue with the additional logic

jblefkove-va (approved):

PR #1008, opened 2022-10-18
Update parent version so we can use ConfigurableValues

KevinCranmer:


data-query-ids-mapping/pom.xml
```diff
@@ -4,7 +4,7 @@
   <parent>
     <groupId>gov.va.api.health</groupId>
     <artifactId>api-starter</artifactId>
-    <version>11.0.8</version>
+    <version>11.0.10</version>
```
KevinCranmer, line 7: `TestValues` was deprecated in favor of `ConfigurationValues` and is causing master builds to fail because of that deprecation. But we can't start using `ConfigurationValues` and let the pipeline auto-upgrade the parent because `ConfigurationValues` was introduced in the same release that `TestValues` was deprecated. So I believe a manual version update is necessary here

jblefkove-va (approved):

PR #1012, opened 2022-10-21
Update library versions

KevinCranmer:


bcastillomartel-va (approved):

data-query/src/test/resources/gov/va/api/health/dataquery/service/controller/medicationorder/datamart-medication-order-v0.json
```diff
@@ -28,7 +28,7 @@
       "additionalInstructions" : "DO NOT TAKE NSAIDS WITH THIS MEDICATION",
       "asNeeded" : false,
       "routeText" : "ORAL",
-      "doseQuantityValue" : 1.0,
+      "doseQuantityValue" : 1,
```
jblefkove-va, line 31: For future reference, it's preferable to avoid editing these datamart payload files, to verify that we are maintaining backwards compatibility; `1.0` here would match `BigDecimal.valueOf(1.0)` in the test

PR #1013, opened 2022-10-25
Add VProcedure Enum values

KevinCranmer:


bcastillomartel-va (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/procedure/Dstu2ProcedureTransformer.java
```diff
@@ -29,10 +29,16 @@ Status status(DatamartProcedure.Status status) {
...
-      return Status.entered_in_error;
+    switch (status) {
+      case cancelled:
+        return Status.entered_in_error;
+      case not_done:
+        return Status.aborted;
+      case unknown:
+        return Status.in_progress;
+      default:
+        return EnumSearcher.of(Procedure.Status.class).find(status.toString());
```
jblefkove-va, line 40: Since there are already three special cases, let's make them all explicit rather than using `EnumSearcher`

jblefkove-va (approved):

PR #1014, opened 2022-11-02
Remove r4 condition sorting

KevinCranmer:
https://vajira.max.gov/browse/API-21459


bcastillomartel-va:
I think we'll also have to update `R4ConditionApi`

KevinCranmer:
> I think we'll also have to update `R4ConditionApi`

Yes good catch. Just updated.

bcastillomartel-va (approved):

bcastillomartel-va:
LGTM

jblefkove-va (approved):

PR #1021, opened 2022-11-28
use fhir-resources 13.0.14

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/practitioner/Dstu2PractitionerTransformer.java
```diff
@@ -164,7 +164,7 @@ List<ContactPoint> telecoms() {
   Practitioner toFhir() {
     return Practitioner.builder()
         .id(datamart.cdwId())
-        .active(BooleanUtils.isTrue(datamart.active()))
```
jblefkove-va, line 167: Previously this was null-pointer protection for `Boolean --> boolean` conversion

PR #1025, opened 2022-12-06
API-22119 observation stored procedure part 2

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -151,16 +163,18 @@ Observation.Bundle searchByPatient(
...
       var spec =
           ObservationRepository.PatientAndDateSpecification.builder().patient(cdwPatient).build();
       var entitiesPage = repository.findAll(spec, page(page, count));
       return bundle(parameters, entitiesPage);
     } else {
       log.info("Using Observation stored procedure");
       int offset = (page - 1) * count + 1;
-      int totalRecords = repository.findObservationsCount(cdwPatient, null, null, null, null);
-      var entities = repository.findObservations(cdwPatient, null, null, null, null, offset, count);
+      var category = "'vital-signs,laboratory,social-history'";
```
jblefkove-va, line 174: @btomlin is this the expected default value for `@Category`?

KevinCranmer (approved):

PR #1031, opened 2022-12-27
Attempt medOrd and medSta readers for MedicationRequest

KevinCranmer:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medicationrequest/R4MedicationRequestController.java
```diff
@@ -145,6 +145,11 @@ private String resourceTypeForId(String publicId) {
       ResourceIdentity resourceIdentity = witnessProtection.toResourceIdentity(publicId);
       return ResourceNameTranslation.get().identityServiceToFhir(resourceIdentity.resource());
     } catch (IdEncoder.BadId e) {
+      if (publicId.contains(":M")) {
```
jblefkove-va, line 148: `contains` or `endsWith`?

PR #1033, opened 2023-01-04
API-18573 Add additional Condition search parameters

bcastillomartel-va:
https://vajira.max.gov/browse/API-18573


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionEntity.java
```diff
@@ -55,6 +55,15 @@ public class ConditionEntity implements CompositeIdDatamartEntity {
   @Column(name = "LastUpdated")
   private Instant lastUpdated;
 
+  @Column(name = "DateUTC")
+  private Instant date;
```
jblefkove-va, line 59: `DateUTC` and `date` are ambiguous names... let's call this `onsetDateTime` to make it clearer what this represents and how we will use it

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionEntity.java
```diff
@@ -55,6 +55,15 @@ public class ConditionEntity implements CompositeIdDatamartEntity {
...
+
+  @Column(name = "ICD10Code")
+  private String icd10Code;
```
jblefkove-va, line 62: Do we also want `ICD9Code`?

bcastillomartel-va: `synthetic-records` only has `ICD10Codes`, should I still add it? @jblefkove-va

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/R4ConditionController.java
```diff
@@ -72,8 +72,11 @@ private VulcanConfiguration<ConditionEntity> configuration() {
                     "clinical-status",
                     this::tokenClinicalStatusIsSupported,
                     this::tokenClinicalStatusSpecification)
+                .string("icd", "icd10Code")
+                .string("snomed", "snomedCode")
                 .values("_id", this::loadCdwId)
                 .dateAsInstant("_lastUpdated", "lastUpdated")
+                .dateAsInstant("_onsetDateTime", "onsetDateTime")
```
jblefkove-va, line 79: `_onsetDateTime` is not the parameter name in the FHIR spec

bcastillomartel-va: are you referring to [src/main/java/gov/va/api/health/r4/api/resources/Condition.java](https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/blob/b222613d60f6a60330f87ae9e92560959bb633c9/us-core-r4/src/main/java/gov/va/api/health/r4/api/resources/Condition.java) ? @jblefkove-va

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/condition/ConditionEntity.java
jblefkove-va: Yes add `icd9Code` as well

jblefkove-va (approved):

PR #1042, opened 2023-01-27
API-23210 medication stored procedure POC

jblefkove-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/medication/Dstu2MedicationController.java
```diff
@@ -70,51 +54,53 @@ private Bundle bundle(
             linkConfig, reports, Medication.Entry::new, Medication.Bundle::new));
   }
 
-  MedicationEntity findById(String publicId) {
-    String cdwId = witnessProtection.toCdwId(publicId);
-    Optional<MedicationEntity> entity =
-        CompositeCdwIds.optionalFromCdwId(cdwId).map(i -> repository.findById(i).orElse(null));
-    if (!entity.isPresent() && !cdwId.endsWith(":L")) {
```
jblefkove-va, line 77: Composite ID migration for this resource was ~18 months ago; it's safe to simplify this method to a single database call

KevinCranmer (approved):

PR #1047, opened 2023-02-09
Update Mitre after Patient41IdBumpAugment

KevinCranmer:
https://vajira.max.gov/browse/API-24082


jblefkove-va:
https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/1046 includes an updated `mitre.mv.db`

PR #1050, opened 2023-02-15
support category search for DocumentReference

bcastillomartel-va:
https://vajira.max.gov/browse/API-23985


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/documentreference/R4DocumentReferenceController.java
```diff
@@ -115,6 +121,38 @@ DocumentReference.Bundle search(HttpServletRequest request) {
...
+   * clinical-note
+   * http://hl7.org/fhir/us/core/ValueSet/us-core-documentreference-category|
+   * http://hl7.org/fhir/us/core/ValueSet/us-core-documentreference-category|clinical-note
+   * </pre>
+   */
+  private boolean tokenCategoryIsSupported(TokenParameter token) {
+    boolean codeIsSupported = token.hasSupportedCode("clinical-note");
+    return (token.hasSupportedSystem(DOCUMENT_REFERENCE_CATEGORY_SYSTEM)
+            && (token.hasAnyCode() || codeIsSupported))
+        || (token.hasAnySystem() && codeIsSupported);
```
jblefkove-va, line 137: Let's do this check first; the logic will then match the order of examples in the javadoc, and I suspect that "supported code with any system" is the most common case

data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4DocumentReferenceApi.java
```diff
@@ -104,6 +104,13 @@ DocumentReference.Bundle documentReferenceSearch(
               name = "type",
               description = "The LOINC code used to identify the document.")
           String type,
+      @Parameter(
+              in = ParameterIn.QUERY,
+              name = "category",
+              description =
+                  "The category the document reference record belongs to."
```
jblefkove-va, line 111: Add a space between the period and the link

jblefkove-va (approved):

PR #1052, opened 2023-02-23
Inferno DocumentReference category.coding.code

bcastillomartel-va:
https://vajira.max.gov/browse/API-23140


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/documentreference/R4DocumentReferenceTransformer.java
```diff
@@ -91,7 +91,8 @@ List<CodeableConcept> categories(String category) {
         .coding(
             Coding.builder()
                 .system("http://hl7.org/fhir/us/core/ValueSet/us-core-documentreference-category")
-                .display(category)
+                .code(category)
+                .display("Clinical Note")
```
jblefkove-va, line 95: Should verify that the `category` is `clinical-note` before hardcoding the display

jblefkove-va (approved):

PR #1061, opened 2023-03-10
API-22438 observation stored procedure part 4

bcastillomartel-va:
https://vajira.max.gov/browse/API-22438


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
```diff
@@ -40,8 +40,9 @@
   @Query(
       value =
           "exec app.usp_Observation "
-              + "@CDWIdNumber = :CDWIdNumber,"
-              + "@CDWIdResourceCode = :CDWIdResourceCode",
+              + "@queryParams = 'CDWIdNumber = :CDWIdNumber "
```
jblefkove-va, line 43: Why `@queryParams`?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -184,15 +184,15 @@ Observation.Bundle searchByPatientAndCategory(
             .build();
     var cdwPatient = witnessProtection.toCdwId(patient);
     var entitiesPage =
-        isBlank(usp)
-            ? repository.findAll(
+        isNotBlank(usp)
```
jblefkove-va, line 187: because `date` parameters are not supported, remove the `usp` header from this method and revert to using the view only

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/ObservationRepository.java
bcastillomartel-va: I thought we should be consistent with `MedicationRepository`

jblefkove-va: The Medication and Observation stored-procedures accept parameters differently; this method is used for reads, so if reads look good in staging with `usp: true`, let's leave it

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -183,16 +183,20 @@ Observation.Bundle searchByPatientAndCategory(
...
-                    .categories(Splitter.on(",").trimResults().splitToList(categoryCsv))
-                    .dates(date)
-                    .build(),
-                page(page, count))
-            : uspObservations(cdwPatient, categoryCsv, null, page, count);
+    Page<ObservationEntity> entitiesPage;
+    try {
+      entitiesPage = uspObservations(cdwPatient, categoryCsv, null, page, count);
+    } catch (Exception e) {
+      log.warn("app.usp_Observation error: {}", Throwables.getRootCause(e).getMessage());
```
jblefkove-va, line 190: The other log statements should match this

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -183,16 +183,20 @@ Observation.Bundle searchByPatientAndCategory(
...
-                ObservationRepository.PatientAndCategoryAndDateSpecification.builder()
-                    .patient(cdwPatient)
-                    .categories(Splitter.on(",").trimResults().splitToList(categoryCsv))
-                    .dates(date)
-                    .build(),
-                page(page, count))
-            : uspObservations(cdwPatient, categoryCsv, null, page, count);
+    Page<ObservationEntity> entitiesPage;
+    try {
+      entitiesPage = uspObservations(cdwPatient, categoryCsv, null, page, count);
```
jblefkove-va, line 188: The stored procedure should only be used if the `date` parameter is null/empty

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -183,16 +183,23 @@ Observation.Bundle searchByPatientAndCategory(
...
-        isBlank(usp)
-            ? repository.findAll(
-                ObservationRepository.PatientAndCategoryAndDateSpecification.builder()
-                    .patient(cdwPatient)
-                    .categories(Splitter.on(",").trimResults().splitToList(categoryCsv))
-                    .dates(date)
-                    .build(),
-                page(page, count))
-            : uspObservations(cdwPatient, categoryCsv, null, page, count);
+    Page<ObservationEntity> entitiesPage;
```
jblefkove-va, line 186: Since the `try` block has its own return statement, it's no longer necessary to declare `entitiesPage` up front, rather, do `var entitiesPage` at the two points where it is used

jblefkove-va (approved):

PR #1070, opened 2023-03-20
Distinguish usp timeout exceptions

bcastillomartel-va:


jblefkove-va:
Closing for now; can reopen if we revisit the stored-procedure approach

PR #1072, opened 2023-03-22
update H2 database

bcastillomartel-va:
[API-24642](https://vajira.max.gov/browse/API-24642)


jblefkove-va:
For future reference, we don't necessarily have to update `mitre.mv.db` every time synthetic data changes. (As long as the local ITs still pass with the old version.) This update could've been deferred until we had a required update.

PR #1089, opened 2023-04-03
Update mitre

KevinCranmer:


jblefkove-va (approved):

jblefkove-va:
For future reference, we don't strictly need to update `mitre.mv.db` until the local ITs break

PR #1091, opened 2023-04-03
API 24874 OAS update MedicationRequest

bcastillomartel-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4MedicationRequestApi.java
```diff
@@ -83,32 +94,31 @@ MedicationRequest.Bundle medicationRequestSearch(
...
               in = ParameterIn.QUERY,
               name = "_id",
               description =
-                  "The logical id of the resource. Once assigned, this value never changes.")
+                  "The MedicationRequest resource FHIR _id. Either the _id or patient "
+                      + "parameter is required. If _id is used, no other parameters can be used.")
           String id,
       @Parameter(
               in = ParameterIn.QUERY,
-              name = "identifier",
```
KevinCranmer, line 97: I don't think we want to remove `identifier`

KevinCranmer (approved):

PR #1095, opened 2023-04-03
API 24874 OAS update Observation

bcastillomartel-va:


KevinCranmer (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4ObservationApi.java
```diff
@@ -81,28 +110,23 @@ Observation.Bundle observationSearch(
...
+                  "The Observation resource FHIR _id. Either the _id or patient parameter "
+                      + "is required. If _id is used, no other parameters can be used.")
           String id,
-      @Parameter(
-              in = ParameterIn.QUERY,
-              name = "identifier",
-              description =
-                  "The logical identifier of the resource. Once assigned, this value "
-                      + "never changes.")
-          String identifier,
```
KevinCranmer, line 99: I don't think we want to be removing this

KevinCranmer: yes we do

PR #1099, opened 2023-04-04
API 24874 OAS update PractitionerRole

bcastillomartel-va:


data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4PractitionerRoleApi.java
```diff
@@ -46,14 +49,18 @@ PractitionerRole practitionerRoleRead(
...
 
   @Operation(
-      summary = "Practitioner Role Search",
+      summary =
+          "Returns practitioner roles by at least one required "
+              + "parameter of _id, practitioner.identifier, or practitioner.name.",
       description =
-          "https://hl7.org/fhir/us/core/STU4/StructureDefinition-us-core-practitionerrole.html",
+          "- A practitioner role is the known role, provider type, specialty,"
+              + " and related data a practitioner can provide for an organization.\n"
```
KevinCranmer, line 61: This one looks to be slightly different than the READ. I'll update it to "●	Practitioner roles are the known roles, provider types, specialties, and related data a practitioner can provide for an organization."

KevinCranmer (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/api/R4PractitionerRoleApi.java
bcastillomartel-va: might want to confirm with Jessica first

PR #1103, opened 2023-04-05
API 25676 Add scopes to well-known endpoint

bcastillomartel-va:
[API-25676](https://vajira.max.gov/browse/API-25676)


jblefkove-va:
update r4 only

jblefkove-va (approved):

PR #1109, opened 2023-04-20
API 23147 update Practitioner.identifier.type.coding system url

bcastillomartel-va:
[API-23147](https://vajira.max.gov/browse/API-23147)


KevinCranmer (commented):
Do we need to approve these with KBS?

PR #1110, opened 2023-04-20
Reverting Recent PRs

KevinCranmer:
This is so we can go to Prod with Patient URL changes without any other changes


KevinCranmer:
Decided it wasn't worth it since we can merge everything a week later

PR #1112, opened 2023-04-21
Update Patient Transformer to be compliant with Inferno tests

KevinCranmer:
https://vajira.max.gov/browse/API-23134

https://inferno.healthit.gov/suites/us_core_v501/lzSCMheva56#2


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/patient/R4PatientTransformer.java
```diff
@@ -251,15 +252,15 @@ static Coding relationshipCoding(DatamartPatient.Contact contact) {
       return null;
     }
     Coding.CodingBuilder builder =
-        Coding.builder().system("http://hl7.org/fhir/patient-contact-relationship");
+        Coding.builder().system("http://hl7.org/fhir/ValueSet/patient-contactrelationship");
     return switch (upperCase(trimToEmpty(contact.type()), Locale.US)) {
-      case "CIVIL GUARDIAN", "VA GUARDIAN" -> builder.code("guardian").display("Guardian").build();
+      case "CIVIL GUARDIAN", "VA GUARDIAN" -> builder.code("O").display("Other").build();
```
KevinCranmer, line 257: Not sure if we want to do this change or not

KevinCranmer: Jessica asked Adam and Adam hasn't responded. Will probably ping again if he doesn't respond till after standup

jblefkove-va: Did we get confirmation?

KevinCranmer: Jay gave me mappings for these values. He also mentioned he wanted to provide a Vista Coding as well, so that's now included.

bcastillomartel-va (approved):

PR #1114, opened 2023-04-25
MedicationRequest, don't use `_requester`

KevinCranmer:
Don't want to merge this until we give consumers a heads up


jblefkove-va:
Does this also require a change to fhir-resources?

KevinCranmer:
> Does this also require a change to fhir-resources?

https://github.com/department-of-veterans-affairs/health-apis-fhir-resources/pull/297

PR #1121, opened 2023-04-27
API 26199 modify MedicationRequest empty bundle

bcastillomartel-va:
[API-26199](https://vajira.max.gov/browse/API-26199)


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/ConfigurableBaseUrlPageLinks.java
```diff
@@ -75,6 +75,10 @@ private <B> List<B> links(AbstractLinkContext<B> context) {
       }
       links.add(context.last());
     }
+    if (context.config.totalRecords() == 0 && links.size() > 1) {
```
jblefkove-va, line 78: Instead of removing the links afterward, let's not add them in the first place if `totalRecords` is 0

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/ConfigurableBaseUrlPageLinks.java
```diff
@@ -63,7 +63,9 @@ private <B> List<B> links(AbstractLinkContext<B> context) {
     List<B> links = new ArrayList<>(5);
     // If recordsPerPage = 0, then only return the self link.
     if (!context.isCountOnly()) {
-      links.add(context.first());
+      if (context.config.totalRecords() > 0) {
```
jblefkove-va, line 66: better, but we can go a step further:
```
boolean hasRecords = context.config().totalRecords() > 0;
if (hasRecords && !context.isCountOnly()) {
  ...
}
```

bcastillomartel-va: would that mean I have to add the `first` and `previous` links under the same if statement?

bcastillomartel-va: what about?
```
boolean hasRecords = context.config().totalRecords() > 0;
    if (!context.isCountOnly()) {
      if (hasRecords) {
        links.add(context.first());
      }
      if (context.hasPrevious()) {
        links.add(context.previous());
      }
    }
    links.add(context.self());
    if (!context.isCountOnly()) {
      if (context.hasNext()) {
        links.add(context.next());
      }
      if (hasRecords) {
        links.add(context.last());
      }
    }
```

jblefkove-va (approved):

PR #1068, opened 2023-03-16
API 24132 restore patient match openapi

bcastillomartel-va:
https://vajira.max.gov/browse/API-24132

PR #1122, opened 2023-05-04
API 26260 OAS update for Patient $match

bcastillomartel-va:
[API-26260](https://vajira.max.gov/browse/API-26260)


jblefkove-va:
Isn't it confusing to have both https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/1068 and this PR?

jblefkove-va (approved):

PR #1125, opened 2023-05-10
Populate DiagnosticReport.code

KevinCranmer:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/R4DiagnosticReportTransformer.java
```diff
@@ -39,7 +40,20 @@ private List<CodeableConcept> category() {
...
+            : (Character) CompositeCdwIds.mapFromCdwId(datamart.cdwId()).get("cdwIdResourceCode");
+    return CodeableConcept.builder()
+        .coding(
+            Coding.builder()
+                .system("http://loinc.org")
+                .code("11502-2")
+                .display("Laboratory report")
+                .build()
+                .asList())
+        .text(cdwIdResourceCode != null && cdwIdResourceCode == 'M' ? "Microbiology" : "Lab Panel")
```
KevinCranmer, line 55: There is an email out to KBS confirming that "Lab Panel" is what we want as the alternative text

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/R4DiagnosticReportTransformer.java
```diff
@@ -39,7 +40,20 @@ private List<CodeableConcept> category() {
   }
 
   private CodeableConcept code() {
-    return CodeableConcept.builder().text("panel").build();
+    Character cdwIdResourceCode =
+        datamart.cdwId() == null
+            ? null
+            : (Character) CompositeCdwIds.mapFromCdwId(datamart.cdwId()).get("cdwIdResourceCode");
```
jblefkove-va, line 46: I'm not sure that this is correct... when the transformer executes, the CDW ID may have already been witness-protected into I2-XXX. Take a look at the appointment resource, which has a similar need to inspect the CDW ID

KevinCranmer: Jay responded and is saying we shouldn't populate the `text` at all... so looks like this is blocked until consensus is reached by KBS

jblefkove-va (approved):

PR #1137, opened 2023-05-24
Bring back Inferno Changes

KevinCranmer:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/R4DiagnosticReportController.java
```diff
@@ -185,14 +185,14 @@ Specification<DiagnosticReportEntity> tokenCategorySpecification(TokenParameter
   }
 
   boolean tokenCodeIsSupported(TokenParameter token) {
-    return (token.hasSupportedCode("panel") && !token.hasExplicitlyNoSystem())
+    return (token.hasSupportedCode("11502-2") && !token.hasExplicitlyNoSystem())
```
KevinCranmer, line 188: Do we need to hardcode here?

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/R4DiagnosticReportController.java
```diff
@@ -185,14 +185,14 @@ Specification<DiagnosticReportEntity> tokenCategorySpecification(TokenParameter
...
         && !token.hasExplicitSystem();
   }
 
   Specification<DiagnosticReportEntity> tokenCodeSpecifications(TokenParameter token) {
     return token
         .behavior()
-        .onAnySystemAndExplicitCode(c -> Specifications.<DiagnosticReportEntity>select("code", c))
+        .onAnySystemAndExplicitCode(c -> Specifications.<DiagnosticReportEntity>select("code", "panel"))
```
KevinCranmer, line 195: Will need DB team to update their records for `code` to have `11502-2` for us to not need this change. Thinking this bandaid is okay until they bring over DiagnosticReport to SQL32

PR #1147, opened 2023-06-08
Diagnostic Report Code search backwards compatibility

KevinCranmer:


data-query/src/main/java/gov/va/api/health/dataquery/service/controller/diagnosticreport/R4DiagnosticReportController.java
```diff
@@ -188,10 +188,8 @@ Specification<DiagnosticReportEntity> tokenCategorySpecification(TokenParameter
...
     return token.hasSupportedCode("11502-2", "panel")
-        && ((token.hasExplicitSystem()
-                && token.hasSupportedSystem(DIAGNOSTIC_REPORT_LOINC_CODE_SYSTEM))
-            || !token.hasExplicitSystem()
-            || token.hasExplicitlyNoSystem());
+        && (!token.hasExplicitSystem()
+            || token.hasSupportedSystem(DIAGNOSTIC_REPORT_LOINC_CODE_SYSTEM));
   }
 
   // TODO remove harcoded search for panel once DB is ready for it
```
bcastillomartel-va, line 195: cruft

KevinCranmer: I think we'll want to remove this line once we actually do remove the search for `panel`. We have to keep it in for now so that we can handle both cases before/after datamart makes the switch

bcastillomartel-va (approved):

jblefkove-va (approved):

PR #1155, opened 2023-06-21
More timing logs on Dstu2Observation.searchByPatientAndCode

KevinCranmer:


bcastillomartel-va (approved):

bcastillomartel-va (approved):

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -168,8 +189,25 @@ Observation.Bundle searchByPatientAndCategory(
...
+    timerAfter = Instant.now().get(ChronoField.MILLI_OF_SECOND);
+    log.info(
+        "Dstu2ObservationController.searchByPatientAndCategory - "
+            + "Building the Parameters map took {}ms",
+        timerAfter - timerBefore);
+    timerBefore = Instant.now().get(ChronoField.MILLI_OF_SECOND);
     var entitiesPage = repository.findAll(spec, page(page, count));
-    return bundle(parameters, entitiesPage);
+    timerAfter = Instant.now().get(ChronoField.MILLI_OF_SECOND);
+    log.info(
```
jhulbert-va, line 200: This is a lot of logging. I was thinking
1. Start clock
2. for each step create a new stopclock var
3. One log that tells us all the information so we don't cruft up the rest of the logs.

data-query/src/main/java/gov/va/api/health/dataquery/service/controller/observation/Dstu2ObservationController.java
```diff
@@ -73,10 +76,20 @@ Observation.Bundle bundle(
...
-            .map(dm -> Dstu2ObservationTransformer.builder().datamart(dm).build().toFhir())
+            .map(
+                dm -> {
+                  int start = Instant.now().get(ChronoField.MILLI_OF_SECOND);
+                  var fhirObservation =
+                      Dstu2ObservationTransformer.builder().datamart(dm).build().toFhir();
+                  int end = Instant.now().get(ChronoField.MILLI_OF_SECOND);
+                  fhirMappingTime.put(fhirObservation.id(), end - start);
+                  return fhirObservation;
+                })
```
jhulbert-va, line 90: I don't think we need to log per record, just the total time transformation took.

bcastillomartel-va: wouldn't this help us find any records that could be taking longer than others?

KevinCranmer: Brian Tomlin thinks there could be specific records causing the problems "We weren't seeing issues on sql52 before we moved it over. I'm wondering if there may still be some issues with the payload and the transformer is having problems with certain records?"

jhulbert-va: If we determine that the latency is occurring in the transformation phase, there are other ways we can test that without creating a massive log statement. I think we need to strike a balance here. We want to get the information we need, but we don't want to make it harder on ourselves by creating spam if some other problem comes along.

jhulbert-va (approved):

bcastillomartel-va (approved):

PR #1156, opened 2023-06-22
Adding support for querying the Procedure_v2 table

KevinCranmer:
The Procedure_v2 table does not have a json payload. Instead it has a table column for each field that was previously in the json payload.

Requests with the header `tableVersion` being `v2` will use the new table. All others will continue to use the json payload table


KevinCranmer:
See https://github.com/department-of-veterans-affairs/health-apis-data-query/pull/1161 instead. Closing this PR

PR #1163, opened 2023-06-28
API-27616 default to procedure flat view for DSTU2

jblefkove-va:
See also https://github.com/department-of-veterans-affairs/health-apis-data-query-deployment/pull/225


KevinCranmer (approved):

KevinCranmer:
Will this break sandbox environments since it's default looking for a sandbox table that doesn't exist?

jblefkove-va:
> Will this break sandbox environments since it's default looking for a sandbox table that doesn't exist?

Yep 😦

KevinCranmer (approved):

